PR #73, opened 2022-08-19
WIP: O11Y-2154: Implement MeterProvider

dustinlessard-wf:

PR #78, opened 2022-09-04
O11Y-2154 : Add MeterProvider classes

dustinlessard-wf:

This library doesn't yet support metrics.  This pull requests introduces the API and SDK implementation for the MeterProvider.  Associated classes like Meter, Counter have only been stubbed out.  Those classes will be developed in additional pull requests.

## Short description of the changes

 - Added MeterProvider API, Noop implementation, SDK implementation
 - Stubbed out all necessary additional classes
 - Opportunistic refactor : alphabetized the exports in api.dart and sdk.dart

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

Please see unit tests

## Checklist:

## Notes:

I messed up my rebase on a previous branch and had a draft pr for it.  That pr is here: https://github.com/Workiva/opentelemetry-dart/pull/73


aviary2-wf:
## Security Insights

## Action Items
---

semveraudit-wf:
### Public API Changes

 Recommendation: **:warning: Minor version bump**


```diff
@@ line 4: package:opentelemetry/src/sdk/common/constants.dart @@
+  String get invalidMeterNameMessage
// Adding a top-level variable is a minor change.
```

```diff
@@ line 7: package:opentelemetry/src/api/metrics/noop/noop_meter.dart @@
+  class NoopMeter implements Meter
// Adding a class is a minor change.


```

```diff
@@ line 6: package:opentelemetry/src/sdk/metrics/metric_options.dart @@
+  class MetricOptions implements MetricOptions
// Adding a class is a minor change.


```

```diff
@@ line 7: package:opentelemetry/src/api/metrics/noop/noop_counter.dart @@
+  class NoopCounter<t> implements Counter<dynamic>
// Adding a class is a minor change.


```

```diff
@@ line 5: package:opentelemetry/src/api/metrics/metric_options.dart @@
+  abstract class MetricOptions
// Adding a class is a minor change.


```
<details>
<summary>Click to see 7 more API Changes</summary><p>

---
```diff
@@ line 7: package:opentelemetry/src/api/metrics/noop/noop_meter_provider.dart @@
+  class NoopMeterProvider implements MeterProvider
// Adding a class is a minor change.


```

```diff
@@ line 8: package:opentelemetry/src/sdk/metrics/meter_provider.dart @@
+  class MeterProvider implements MeterProvider
// Adding a class is a minor change.


```

```diff
@@ line 7: package:opentelemetry/src/sdk/metrics/meter.dart @@
+  class Meter implements Meter
// Adding a class is a minor change.


```

```diff
@@ line 6: package:opentelemetry/src/api/metrics/meter.dart @@
+  abstract class Meter
// Adding a class is a minor change.


```

```diff
@@ line 6: package:opentelemetry/src/api/metrics/counter.dart @@
+  abstract class Counter<t>
// Adding a class is a minor change.


```

```diff
@@ line 6: package:opentelemetry/src/sdk/metrics/counter.dart @@
+  class Counter<t> implements Counter<dynamic>
// Adding a class is a minor change.


```

```diff
@@ line 7: package:opentelemetry/src/api/metrics/meter_provider.dart @@
+  abstract class MeterProvider
// Adding a class is a minor change.


```
---

</p></details>

Showing results for 438ff9910057f0e9b3d8d435cecaed84790cdf71
[Browse public API](https://semver-audit.wk-dev.wdesk.org/Workiva/opentelemetry-dart/438ff9910057f0e9b3d8d435cecaed84790cdf71).

lib/api.dart
```diff
@@ -2,24 +2,31 @@
...
 export 'src/api/context/context.dart' show Context;
 export 'src/api/exporters/span_exporter.dart' show SpanExporter;
 export 'src/api/instrumentation_library.dart' show InstrumentationLibrary;
+export 'src/api/metrics/counter.dart' show Counter;
+export 'src/api/metrics/meter_provider.dart' show MeterProvider;
+export 'src/api/metrics/meter.dart' show Meter;
+export 'src/api/metrics/metric_options.dart' show MetricOptions;
+export 'src/api/metrics/noop/noop_counter.dart' show NoopCounter;
+export 'src/api/metrics/noop/noop_meter_provider.dart' show NoopMeterProvider;
+export 'src/api/metrics/noop/noop_meter.dart' show NoopMeter;
```
blakeroberts-wk, line 16: Is it important to export these?

lib/src/api/metrics/noop/noop_meter.dart
```diff
@@ -0,0 +1,11 @@
+import 'package:opentelemetry/api.dart';
+
+/// A no-op instance of a [Meter]]
+class NoopMeter implements Meter {
+  static final _noopCounter = NoopCounter<int>();
```
blakeroberts-wk, line 5: Would it ever make sense to instantiate a non-int `NoopCounter`? Could instead the `NoopCounter` simply `implements Counter<int>`?

lib/src/sdk/metrics/counter.dart
```diff
@@ -0,0 +1,9 @@
+import 'package:opentelemetry/api.dart' as api;
+
+class Counter<t> implements api.Counter {
+  @override
+  void add(dynamic value,
```
blakeroberts-wk, line 5: Can `value` not take on type `t`? If we use `dynamic`, won't the implementation be forced to type check manually?

blakeroberts-wk: Also, is it possible to limit `t` to numbers only? I.e. can we do `class Counter<t implements num> implements api.Counter`? If yes, perhaps that constraint should be placed on `api.Counter`. Though I'm guessing this wouldn't support `BigInt` or `fixnum`?

lib/src/sdk/metrics/meter_provider.dart
```diff
@@ -0,0 +1,34 @@
...
+class MeterProvider implements api.MeterProvider {
+  final _meters = <String, api.Meter>{};
+  final _logger = Logger('opentelemetry.sdk.metrics.meterprovider');
+
+  @override
+  sdk.Meter get(String instrumentationScopeName,
+      {String instrumentationVersion = '',
+      String schemaUrl = '',
+      Map<String, String> attributes = const {}}) {
+    if (instrumentationScopeName == null || instrumentationScopeName == '') {
```
blakeroberts-wk, line 14: I think we should throw an error here, null or empty values aren't allowed: https://opentelemetry.io/docs/reference/specification/metrics/api/#instrument-naming-rule

lib/api.dart
dustinlessard-wf: Good point.  Will fix.

lib/src/sdk/metrics/meter_provider.dart
dustinlessard-wf: I originally wrote this code to throw an ArgumentError but then read [this](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#get-a-meter)

> In case an invalid name (null or empty string) is specified, a working Meter implementation MUST be returned as a fallback rather than returning null or throwing an exception, its name property SHOULD keep the original invalid value, and a message reporting that the specified value is invalid SHOULD be logged.

I've adjusted the code to instead produce an appropriate log, complete with the stacktrace so that one can track down the offending code.

lib/src/api/metrics/noop/noop_meter.dart
dustinlessard-wf: Good catch - in this case, it doesn't make any difference.  I've changed it as suggested.

lib/src/sdk/metrics/counter.dart
dustinlessard-wf: Screen Shot 2022-09-07 at 4 32 14 PM
I tried this, but it doesn't like the use of 't' here.

dustinlessard-wf: I think it's better to avoid the assumption that the only types of numbers we will want to use are those that implement num.

blakeroberts-wk: I think you may need to first update the class declaration to `class Counter<t> implements api.Counter<t>`. Also, nit: I'd use `T` since it is referring to a class/type which are usually capitalized

dustinlessard-wf: You're right- THANKS!

lib/src/api/metrics/noop/noop_counter.dart
```diff
@@ -0,0 +1,7 @@
+import 'package:opentelemetry/api.dart';
+
+/// A no-op instance of a [Counter]
+class NoopCounter implements Counter<int> {
+  @override
+  void add(dynamic value, {List<Attribute> attributes, Context context}) {}
```
blakeroberts-wk, line 6:
```suggestion
  void add(int value, {List<Attribute> attributes, Context context}) {}
```

lib/src/sdk/common/constants.dart
```diff
@@ -0,0 +1 @@
+const invalidMeterNameMessage = 'Invalid Meter Name';
```
blakeroberts-wk, line 1: nit: I don't like constants files. Just makes me have to click more links to find values. tbh, if the value is used once, I'd inline it. No reason to have a constant. If the desire is to `if (err == invalidMeterNameMessage)`, I'd make the const the actual error/exception being thrown, not a string. At a minimum, I'd throw this in the meter file

lib/src/sdk/metrics/meter.dart
```diff
@@ -0,0 +1,9 @@
+import 'package:opentelemetry/api.dart' as api;
+import 'package:opentelemetry/sdk.dart' as sdk;
+
+class Meter implements api.Meter {
+  @override
+  api.Counter createCounter<t>(String name, {api.MetricOptions options}) {
+    return sdk.Counter<t>();
```
blakeroberts-wk, line 7:
```suggestion
  api.Counter createCounter<T>(String name, {api.MetricOptions options}) {
    return sdk.Counter<T>();
```

lib/src/sdk/metrics/meter_provider.dart
```diff
@@ -0,0 +1,34 @@
...
+      instrumentationScopeName = '';
+      //throw and catch Argument error so that we capture a stacktrace,
+      //identifying the caller
+      try {
+        throw ArgumentError(sdk.invalidMeterNameMessage);
+      } catch (e, stacktrace) {
+        _logger.warning(sdk.invalidMeterNameMessage, e, stacktrace);
+      }
```
blakeroberts-wk, line 22: Is it possible to use https://api.dart.dev/stable/2.3.0/dart-core/StackTrace/current.html and avoid needing to throw/catch?

lib/src/api/metrics/meter.dart
```diff
@@ -0,0 +1,13 @@
...
+abstract class Meter {
+  /// Creates a new [Counter] metric. Generally, this kind of metric is used
+  /// when the value is a quantity, the sum is of primary interest, and the
+  /// event count and value distribution are not of primary interest.
+  ///
+  /// [name] The name of the [Counter] See https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#instrument-naming-rule"
+  /// [options] Options needed for metric creation. Pattern borrowed from
+  /// opentelemetry-js. This is optional.
+  ///
+  Counter createCounter<t>(String name, {MetricOptions options});
```
blakeroberts-wk, line 12:
```suggestion
  Counter createCounter<T>(String name, {MetricOptions options});
```

lib/src/api/metrics/counter.dart
```diff
@@ -0,0 +1,11 @@
+import 'package:opentelemetry/api.dart';
+
+abstract class Counter<T> {
```
blakeroberts-wk, line 3:
```suggestion
abstract class Counter<T extends num> {
```

lib/src/sdk/common/constants.dart
dustinlessard-wf: I have no strong opinion here. I'll make this change.

lib/src/sdk/metrics/meter.dart
```diff
@@ -0,0 +1,11 @@
+import 'package:opentelemetry/api.dart' as api;
+import 'package:opentelemetry/sdk.dart' as sdk;
+
+class Meter implements api.Meter {
+  static const invalidMeterNameMessage = 'Invalid Meter Name';
```
blakeroberts-wk, line 5: I guess by meter file I meant meter_provider.dart. And can we make it private?

dustinlessard-wf: LOL ... I was thinking the same thing, but then didn't have a strong opinion as there's a case to be made for housing it in either place.

blakeroberts-wk (dismissed):

blakeroberts-wk (dismissed):

dustinlessard-wf:
QA +1, unit tests cover the limited concrete implementation (MeterProvider)

lib/src/api/metrics/meter.dart
```diff
@@ -0,0 +1,13 @@
+import 'package:opentelemetry/api.dart';
+
+abstract class Meter {
+  /// Creates a new [Counter] metric. Generally, this kind of metric is used
```
evanweible-wf, line 4: The spec refers to these as "instruments" rather than "metrics" - should we align with that?

lib/src/api/metrics/meter.dart
```diff
@@ -0,0 +1,13 @@
...
+  /// when the value is a quantity, the sum is of primary interest, and the
+  /// event count and value distribution are not of primary interest.
+  ///
+  /// [name] The name of the [Counter] See https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#instrument-naming-rule"
+  /// [options] Options needed for metric creation. Pattern borrowed from
+  /// opentelemetry-js. This is optional.
```
evanweible-wf, line 10: #nit I'm not sure the part about this pattern being borrowed from opentelemetry-js is necessary here. The doc comment should be written for the consumer, and this feels like a note intended for the maintainers.

lib/src/sdk/metrics/metric_options.dart
```diff
@@ -0,0 +1,13 @@
...
+  @override
+  void setDescription(String description) {
+    // TODO: implement setDescription https://github.com/Workiva/opentelemetry-dart/issues/76
+  }
+
+  @override
+  void setUnits(String units) {
+    // TODO: implement setUnits https://github.com/Workiva/opentelemetry-dart/issues/77
+  }
+}
```
evanweible-wf, line 13: I know the options object pattern is common in JS, and that opentelemetry-js uses it for instrument creation, but I'm not sure it makes sense in Dart for a few reasons:
- Dart supports positional, named optional, and named required parameters, which makes it possible to specify all of the "options" directly as parameters on the creation method rather than fields on an options object. 
- Object construction is (slightly) more verbose in Dart than it is in JS. In JS, you can create these options objects very tersely - for example, if you have same-named variables, it could be as simple as `{ description, units }`. The same in Dart would require at least something like: `MetricOptions(description: description, units: units)`.
- The value of the "builder" pattern that is stubbed out above is not clear to me. If absolutely necessary, callers could always build up the necessary information incrementally using local variables rather than fields on a class.

The one value I do see being provided by this class is that it can be shared across all instrument types. But IMO, I think that really only benefits the maintainers and I don't think that's enough to justify its existence. As a consumer of the `Meter` interface, I would rather be able to supply the parameters directly to the method rather than needing to create a separate object. The consumer will still get the benefit of consistency of parameters across all of the creation methods on the `Meter` interface with the object.

```
  Counter<T> createCounter<T>(String name, {String unit, String description});
```

lib/src/api/metrics/meter.dart
```diff
@@ -0,0 +1,13 @@
...
+  ///
+  Counter createCounter<T>(String name, {MetricOptions options});
```
evanweible-wf, line 12: For the generic to work, I think this will need to be:
```suggestion
  Counter<T> createCounter<T>(String name, {MetricOptions options});
```

lib/src/api/metrics/counter.dart
```diff
@@ -0,0 +1,11 @@
+import 'package:opentelemetry/api.dart';
+
+abstract class Counter<T extends num> {
+  /// Records a value with a set of attributes.
+  ///
+  /// [value] The increment amount. MUST be non-negative.
+  /// [attributes] A set of attributes to associate with the value.
+  /// [context] The explicit context to associate with this measurement.
+  ///
```
evanweible-wf, line 9:
```suggestion
```

lib/src/api/metrics/counter.dart
```diff
@@ -0,0 +1,11 @@
...
+  void add(T value, {List<Attribute> attributes, Context context});
```
evanweible-wf, line 10: Just out of curiosity, why are these attributes typed as `List<Attributes>` while the attributes used by `MeterProvider.get(...)` are typed as `Map<String, String>`? Are they actually different or is this just a different way to represent the same data?

lib/src/api/metrics/noop/noop_counter.dart
```diff
@@ -0,0 +1,7 @@
+import 'package:opentelemetry/api.dart';
+
+/// A no-op instance of a [Counter]
+class NoopCounter implements Counter<int> {
+  @override
+  void add(int value, {List<Attribute> attributes, Context context}) {}
+}
```
evanweible-wf, line 7: #nit - I'm not sure if it matters, but if this is already a no-op, is there any reason to clamp down the type to `int` rather than letting it default to the more permissive `num`? Only reason I ask is because I'm envisioning a scenario where someone creates a counter and (based purely on app configuration) gets a `NoopCounter` and then tries to call `.add()` with a double. That would probably throw at runtime rather than being a no-op.

lib/src/api/metrics/noop/noop_meter.dart
```diff
@@ -0,0 +1,12 @@
+import 'package:opentelemetry/api.dart';
+import 'package:opentelemetry/src/api/metrics/noop/noop_counter.dart';
+
+/// A no-op instance of a [Meter]]
+class NoopMeter implements Meter {
+  static final _noopCounter = NoopCounter();
+
+  @override
+  Counter createCounter<t>(String name, {MetricOptions options}) {
```
evanweible-wf, line 9: Building off the above comment, I think this might need to be typed such that it always returns a `Counter<num>` to avoid a runtime error.. I'd need to test it out to know for sure.

lib/src/sdk/metrics/meter_provider.dart
```diff
@@ -0,0 +1,32 @@
...
+  final _logger = Logger('opentelemetry.sdk.metrics.meterprovider');
+
+  @visibleForTesting
+  static const invalidMeterNameMessage = 'Invalid Meter Name';
+
+  @override
+  sdk.Meter get(String instrumentationScopeName,
+      {String instrumentationVersion = '',
+      String schemaUrl = '',
+      Map<String, String> attributes = const {}}) {
```
evanweible-wf, line 17: FYI - there's an annoying quirk of pre-null-safe Dart where optional parameters with default values can still end up with `null` values. If anyone were to call this like so:

```
String schemaUrl;
meterProvider.get('foo', schemaUrl: schemaUrl);
```

Then in here, `schemaUrl` would actually be null instead of the default value specified here. I don't think that will be a problem for implementing this method, but something to be aware of.

test/unit/api/metrics/noop_meter_provider_test.dart
```diff
@@ -0,0 +1,48 @@
...
+
+void main() {
+  group('MeterProvider:', () {
+    setUp(() {});
+
+    test('MeterProvider.get with name returns inert instance of Meter', () {
+      final provider = NoopMeterProvider();
+      expect(provider, isA<api.MeterProvider>());
+      final meter = provider.get('testname');
+      expect(meter, isA<api.Meter>());
```
evanweible-wf, line 14: #nit - `isA<Type>()` matchers used like this don't provide much value since the Dart type system is already guaranteeing this for you since `NoopMeterProvider` implements `MeterProvider`. You wouldn't be able to call `.get(...)` on the provider or `.createCounter` on the meter if that weren't the case, so you can safely omit these checks.

Really the only time these are needed are if you're testing for a _more_ specific type, like a function that returns `num` and you want to verify that in this case it's actually an `int`.

test/unit/sdk/metrics/meter_provider_test.dart
```diff
@@ -0,0 +1,159 @@
...
+
+import 'package:test/test.dart';
+
+void main() {
+  group('MeterProvider:', () {
+    setUp(() {
+      Logger.root.level = Level.ALL; // defaults to Level.INFO
+      Logger.root.onRecord.listen((record) {
+        print('${record.level.name}: ${record.time}: ${record.message}');
+      });
```
evanweible-wf, line 17: #fyi there's a really useful [`printOnFailure` utility](https://api.flutter.dev/flutter/test_api.scaffolding/printOnFailure.html) provided by the test package that comes in handy for things like this:

```suggestion
      Logger.root.level = Level.ALL; // defaults to Level.INFO
      Logger.root.onRecord.listen((record) {
        printOnFailure('${record.level.name}: ${record.time}: ${record.message}');
      });
```

This way, you'll only see the logs from a test if it fails without having to always clutter up the test run output.

Also, depending on what you're going for, the default `toString()` representation provided by `LogRecord` looks like so: `'[${level.name}] $loggerName: $message'` - so you might even be able to simplify this to:

```suggestion
      Logger.root.level = Level.ALL; // defaults to Level.INFO
      Logger.root.onRecord.listen(printOnFailure);
      });
```

test/unit/sdk/metrics/meter_provider_test.dart
```diff
@@ -0,0 +1,159 @@
...
+
+    test(
+        'failing to provide a name when getting a meter results in a functional'
+        'meter and logs a message', () {
+      Logger.root.onRecord.listen(expectAsync1((record) {
+        expect(record.stackTrace, isNotNull);
+        expect(
+            record.message, equals(sdk.MeterProvider.invalidMeterNameMessage));
+        expect(record.level, equals(Level.WARNING));
+      }));
```
evanweible-wf, line 28: [Stream matchers](https://pub.dev/packages/test#stream-matchers) are going to be really helpful for tests like this.  For example:
```suggestion
      expect(
          Logger.root.onRecord,
          emits(isA<LogRecord>()
              .having(
                (r) => r.message,
                'message',
                equals(sdk.MeterProvider.invalidMeterNameMessage),
              )
              .having(
                (r) => r.level,
                'level',
                equals(Level.WARNING),
              )));
```

There a couple of benefits to this approach:
- Stream matchers in combination with the `isA<Type>().having(...)` provides helpful error messages:
```
Expected: should emit an event that <Instance of 'LogRecord'> with `message`: 'Invalid Meter Name' and `level`: Level:<INFO>
  Actual: <Instance of '_BroadcastStream<LogRecord>'>
   Which: emitted • [WARNING] opentelemetry.sdk.metrics.meterprovider: Invalid Meter Name
            which emitted an event that has `level` with value Level:<WARNING>
```
- You can use more flexible stream matchers like `emitsThrough(...)` to be more resilient in scenarios where you may not have full control over what is being logged.

test/unit/sdk/metrics/meter_provider_test.dart
```diff
@@ -0,0 +1,159 @@
...
+      expect(meter, isNotNull);
+    });
+
+    test('getting a meter by same name will return the same instance', () {
+      const meterName = 'meterA';
+      final meterProvider = sdk.MeterProvider();
+      final meterA = meterProvider.get(meterName);
+      final meterB = meterProvider.get(meterName);
+
+      expect(identical(meterA, meterB), true);
```
evanweible-wf, line 40: #nit this is probably fine since it's at least verifying that the code is doing what it should as written, but the spec doesn't actually require this behavior:

> It is unspecified whether or under which conditions the same or different Meter instances are returned from this function when the same (name,version,schema_url,attributes) parameters are used.

lib/src/sdk/metrics/meter.dart
```diff
@@ -0,0 +1,9 @@
+import 'package:opentelemetry/api.dart' as api;
+import 'package:opentelemetry/sdk.dart' as sdk;
+
+class Meter implements api.Meter {
+  @override
+  api.Counter createCounter<T>(String name, {api.MetricOptions options}) {
```
evanweible-wf, line 6:
```suggestion
  api.Counter<T> createCounter<T>(String name, {api.MetricOptions options}) {
```

lib/src/api/metrics/meter_provider.dart
```diff
@@ -0,0 +1,18 @@
...
+abstract class MeterProvider {
+  ///
+  /// Gets or creates a named [Meter] instance.
+  ///
+  /// [instrumentationScopeName] A name uniquely identifying the instrumentation scope, such as
+  /// the instrumentation library, package, or fully qualified class name. Must not be null.
+  /// Please see https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/glossary.md#instrumentation-scope for more details
+  /// [instrumentationVersion] A version associated with the [instrumentationScopeName]
+  /// [schemaUrl] The URL of the OpenTelemetry schema being used by this instrumentation library.
+  /// [attributes] Specifies the instrumentation scope attributes to associate with emitted telemetry.
```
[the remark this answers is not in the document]
dustinlessard-wf: Adjusted.  I left the link to the additional documentation however as I believe this is helpful when a caller is getting or creating a MeterProvider

lib/src/api/metrics/meter_provider.dart
```diff
@@ -0,0 +1,18 @@
...
+  Meter get(String instrumentationScopeName,
+      {String instrumentationVersion = '',
```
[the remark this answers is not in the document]
dustinlessard-wf: Agreed.

test/unit/api/metrics/noop_meter_provider_test.dart
dustinlessard-wf: Good point.  Thanks!

test/unit/sdk/metrics/meter_provider_test.dart
dustinlessard-wf: Agreed, I believe we should adopt this approach now, to support this part of the spec:
> Implementations MUST NOT require users to repeatedly obtain a Meter with the same identity to pick up configuration changes. This can be achieved either by allowing to work with an outdated configuration or by ensuring that new configuration applies also to previously returned Meters.

My thinking is that if two places of code have a reference to the same instance of a Meter, then should the configuration of that meter be changed, we can be confident the change will be used by both references.

test/unit/sdk/metrics/meter_provider_test.dart
dustinlessard-wf: I tried the shortened version but it didn't like the type mismatch so I went with your first suggestion.

lib/src/sdk/metrics/meter_provider.dart
```diff
@@ -0,0 +1,32 @@
...
+    if (instrumentationScopeName == null || instrumentationScopeName == '') {
+      instrumentationScopeName = '';
+      _logger.warning(invalidMeterNameMessage, '', StackTrace.current);
+    }
+    final key = instrumentationScopeName +
+        instrumentationVersion +
+        schemaUrl +
+        attributes.toString();
```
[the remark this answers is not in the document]
dustinlessard-wf: This is way better than my naive implementation! THANKS

lib/src/api/metrics/meter.dart
dustinlessard-wf: Followed previous advice and adopted a more narrative style, cut this comment down quite a bit.

lib/src/sdk/metrics/metric_options.dart
dustinlessard-wf: I debated this too and considered a few different approaches:
- Dart's named optional parameters
- Configuration object pattern
- Builder pattern

TBH, for me, it was a toss up between the configuration object pattern and using Dart's native named parameters.  I appreciate you expressing your preference.  I'll switch to named optional parameters.

lib/src/api/metrics/noop/noop_counter.dart
```diff
@@ -0,0 +1,7 @@
+import 'package:opentelemetry/api.dart';
+
+/// A no-op instance of a [Counter]
+class NoopCounter<T> extends Counter<num> {
```
dustinlessard-wf, line 4: I could use your expertise @evanweible-wf @blakeroberts-wk .  I can't seem to get the definition of the NoopCounter to jive with the Counter class, and am having trouble squaring this up with returning a NoopCounter from the Meter class: https://github.com/Workiva/opentelemetry-dart/pull/78/files#diff-bea4968f0e49128097b068cd99c2e557336e70ed230a42c9d7d44e6ca944ee31R11

dustinlessard-wf: I keep getting static analysis errors.
Screen Shot 2022-09-12 at 11 07 56 AM
Screen Shot 2022-09-12 at 11 07 46 AM

blakeroberts-wk: It might be that you want `class NoopCounter extends Counter<num>` and `api.Counter<T> createCounter<T extends num>(String name, ...)`?

dustinlessard-wf: That doesn't work either. :(
Screen Shot 2022-09-12 at 1 42 11 PM

dustinlessard-wf:
@evanweible-wf @blakeroberts-wk , Please don't re-review at this time. I have some comments to address first. THANKS!

lib/src/api/metrics/noop/noop_counter.dart
dustinlessard-wf: I got it working.

lib/src/api/metrics/counter.dart
dustinlessard-wf: This is an oversight.

Here's what the two different usages look like:

` final meter = provider.get('testname',
          version: 'version',
          schemaUrl: 'url',
          attributes: {'http.method': 'post', 'http.scheme': 'http'});`
vs
` meter.createCounter('test').add(1, attributes: [
        api.Attribute.fromString('http.method', 'http.scheme'),
        api.Attribute.fromString('http.scheme', 'http')
      ]);`

I think the use of a general purpose map results in less code to write without impacting readability.  I don't see the benefit of adopting the current `api.Attribute.fromString` pattern.
Thoughts? @evanweible-wf @blakeroberts-wk 
Thoughts?

lib/src/api/metrics/noop/noop_counter.dart
dustinlessard-wf: Good point - changed.

dustinlessard-wf:
This comment is collapsed but could benefit from your explicit review: https://github.com/Workiva/opentelemetry-dart/pull/78#discussion_r967491560

dustinlessard-wf:
Ready for re-review!

lib/src/api/metrics/counter.dart
evanweible-wf: Yeah I'd agree, being able to specify attributes as a simple map is much easier and probably more intuitive for consumers. Unless there's a good reason for requiring a specific Attribute type, I'd vote for the map.

lib/src/api/metrics/meter.dart
```diff
@@ -0,0 +1,11 @@
+import 'package:opentelemetry/api.dart';
+
+abstract class Meter {
+  /// Creates a new [Counter] instrument named [name]. Additional details about
+  /// this metric can be captured in [description] and units can be specified in
+  /// [units].
+  ///
+  /// See https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#instrument-naming-rule"
```
evanweible-wf, line 8:
```suggestion
  /// See https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#instrument-naming-rule
```

lib/src/api/metrics/meter_provider.dart
```diff
@@ -0,0 +1,19 @@
...
+abstract class MeterProvider {
+  /// Gets or creates a [Meter] instance.
+  ///
+  /// The meter is identified by the combination of [name], [version],
+  /// [schemaUrl] and [attributes]. The [name] SHOULD uniquely identify the
+  /// instrumentation scope, such as the instrumentation library
+  /// (e.g. io.opentelemetry.contrib.mongodb), package, module or class name.
+  /// The [version] specifies the version of the instrumentation scope if the
+  /// scope has a version (e.g. a library version). The [schemaUrl] identifies
+  /// the schema this provider ahdheres to.  The [attributes] specifies
```
evanweible-wf, line 13:
```suggestion
  /// the schema this provider adheres to.  The [attributes] specifies
```

lib/src/api/metrics/meter_provider.dart
```diff
@@ -0,0 +1,19 @@
...
+  /// attributes to associate with emitted telemetry
```
evanweible-wf, line 14:
```suggestion
  /// attributes to associate with emitted telemetry.
```

lib/src/api/metrics/noop/noop_meter.dart
```diff
@@ -0,0 +1,11 @@
+import 'package:opentelemetry/api.dart';
+import 'package:opentelemetry/src/api/metrics/noop/noop_counter.dart';
+
+/// A no-op instance of a [Meter]]
```
evanweible-wf, line 4:
```suggestion
/// A no-op instance of a [Meter]
```

evanweible-wf (commented):
Only a few nits and then the discussion around typing of attributes, otherwise LGTM!

lib/src/api/metrics/counter.dart
blakeroberts-wk: things to keep in mind:
- otel only supports so many attribute key types, so strongly typing these is advised.
- duplicates are not allowed (hence internal map representation is good), but otlp formats attributes as a list (when serialized to protobuf json or binary)
- usually, semantic conventions should be followed and therefore the key used should be of a constant defined within the otel library (e.g. https://github.com/Workiva/messaging-sdk/blob/master/lib/java/sdk/src/main/java/com/workiva/messaging_sdk/tracing/ServerTracingMiddleware.java#L85)

blakeroberts-wk: Here is the specification for attributes: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/README.md#attribute.

evanweible-wf: That's helpful, thanks! The restrictions on the keys seem manageable with a Dart `Map`, but this part:

> - The attribute value is either:
>   - A primitive type: string, boolean, double precision floating point (IEEE 754-1985) or signed 64 bit integer.
>   - An array of primitive type values. The array MUST be homogeneous, i.e., it MUST NOT contain values of different types.

\- makes me lean towards using an `Attribute` class. That way we could provide strongly-typed constructors/factories for each of those supported value types (string, bool, num, or a homogeneous list of those types). Something like:

```
Attribute.ofString(SemanticAttributes.FOO, 'foo')
Attribute.ofNum(SemanticAttributes.BAR, 2.0)
Attribute.ofBools(SemanticAttributes.BAZ, [true, false, false])
```

dustinlessard-wf: From what I gather from that readme, 
> Attributes, which belong to Metrics, are exempt from the limits described above at this time, as discussed in [Metrics Attribute Limits](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#attribute-limits).

That being said, I'm leaning towards using the existing attribute class because it's likely that the type restrictions on metrics will eventually align with those of spans.

lib/src/sdk/metrics/metric_options.dart
evanweible-wf: Will this happen in a follow up?

evanweible-wf (approved):

lib/src/api/metrics/counter.dart
blakeroberts-wk: Those limits I believe refer to limits of number of attributes and length of keys/values of attributes. I believe the types of attributes will remain the same. This is at least how the otlp protobuf definitions are written: https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto#L341

dustinlessard-wf:
I think we should hide these changes from the public api for this package.  This pr doesn't introduce any functionality that a consumer can yet use and may change once we get deeper into the implementation for the various instruments.  I've grouped the new api bits into /lib/src/experimental_api.dart and the sdk bits into /lib/src/experimental_sdk.dart.

evanweible-wf (approved):

lib/src/api/metrics/noop/noop_counter.dart
```diff
@@ -0,0 +1,8 @@
+import 'package:opentelemetry/api.dart';
+import 'package:opentelemetry/src/experimental_api.dart';
+
+/// A no-op instance of a [Counter]
+class NoopCounter<T extends num> extends Counter<T> {
```
blakeroberts-wk, line 5: Can we do this?
```suggestion
class NoopCounter extends Counter<num> {
```

pubspec.yaml
```diff
@@ -10,6 +10,7 @@ dependencies:
   http: ^0.12.0
   logging: '>=0.11.4 <2.0.0'
   protobuf: '>=1.1.0 <3.0.0'
+  quiver: '>=2.1.5 <4.0.0'
```
blakeroberts-wk, line 13: Does it make sense to copy the `hash4` implementation to avoid adding a new dependency?

blakeroberts-wk (approved):
two nits, nothing blocking

pubspec.yaml
blakeroberts-wk: Here's the code we need:
```dart
/// Generates a hash code for four objects.
int hash4(a, b, c, d) => _finish(_combine(
    _combine(_combine(_combine(0, a.hashCode), b.hashCode), c.hashCode),
    d.hashCode));

// Jenkins hash functions

int _combine(int hash, int value) {
  hash = 0x1fffffff & (hash + value);
  hash = 0x1fffffff & (hash + ((0x0007ffff & hash) << 10));
  return hash ^ (hash >> 6);
}

int _finish(int hash) {
  hash = 0x1fffffff & (hash + ((0x03ffffff & hash) << 3));
  hash = hash ^ (hash >> 11);
  return 0x1fffffff & (hash + ((0x00003fff & hash) << 15));
}
```
https://github.com/google/quiver-dart/blob/master/lib/src/core/hash.dart#L26-L43

evanweible-wf: I'd be fine with that, we just need to provide attribution via a code comment.

lib/src/api/metrics/noop/noop_counter.dart
dustinlessard-wf: We cannot.  That would also require changing the 'add' function definition which would then no longer align with the Counter abstract class.

pubspec.yaml
dustinlessard-wf: I'm going to capture this as a maybe-todo.  I don't see that the pros of doing this outweigh the cons and I'd like to discuss it further before we do this.  If we decide to do this, it wouldn't impact the public api, so I think it's ok to followup on this one.

dustinlessard-wf:
QA +1 , pulled into a Flutter test project and verified that a MeterProvider can be constructed.

dustinlessard-wf:
@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #90, opened 2022-11-05
O11Y-2256 : Resource can be passed into MeterProvider

dustinlessard-wf:

It should be possible to associate a Resource with a MeterProvider so that all derivative instruments share the resource.

These changes allow a resource to be passed into the MeterProvider constructor and further build on the effort to bring metrics support to this library

## How Has This Been Tested?

- Please see unit-test

## Checklist:

@Workiva/observability

### DIFF
--- a/lib/src/experimental_sdk.dart
+++ b/lib/src/experimental_sdk.dart
@@ -1,3 +1,7 @@
 export 'sdk/metrics/counter.dart' show Counter;
 export 'sdk/metrics/meter_provider.dart' show MeterProvider;
 export 'sdk/metrics/meter.dart' show Meter;
+export 'sdk/metrics/state/meter_shared_state.dart' show MeterSharedState;
+export 'sdk/metrics/state/meter_provider_shared_state.dart'
+    show MeterProviderSharedState;
+export 'sdk/resource/resource.dart' show Resource;
--- a/lib/src/experimental_sdk.dart
+++ b/lib/src/experimental_sdk.dart
@@ -6,7 +6,4 @@ import 'package:meta/meta.dart';
 export 'sdk/metrics/counter.dart' show Counter;
 export 'sdk/metrics/meter_provider.dart' show MeterProvider;
 export 'sdk/metrics/meter.dart' show Meter;
-export 'sdk/metrics/state/meter_shared_state.dart' show MeterSharedState;
-export 'sdk/metrics/state/meter_provider_shared_state.dart'
-    show MeterProviderSharedState;
 export 'sdk/resource/resource.dart' show Resource;
--- a/lib/src/experimental_sdk.dart
+++ b/lib/src/experimental_sdk.dart
@@ -1,3 +1,8 @@
+@experimental
+library experimental_sdk;
+
+import 'package:meta/meta.dart';
+
 export 'sdk/metrics/counter.dart' show Counter;
 export 'sdk/metrics/meter_provider.dart' show MeterProvider;
 export 'sdk/metrics/meter.dart' show Meter;
--- a/lib/src/sdk/common/instrumentation_scope.dart
+++ b/lib/src/sdk/common/instrumentation_scope.dart
@@ -1,9 +1,13 @@
+import 'package:opentelemetry/api.dart' as api;
+
 class InstrumentationScope {
   final String _name;
   final String _version;
   final String _schemaUrl;
+  final List<api.Attribute> _attributes;
 
-  InstrumentationScope(this._name, this._version, this._schemaUrl);
+  InstrumentationScope(
+      this._name, this._version, this._schemaUrl, this._attributes);
 
   String get name {
     return _name;
@@ -15,5 +19,9 @@ class InstrumentationScope {
 
   String get schemaUrl {
     return _schemaUrl;
+  }
+
+  List<api.Attribute> get attributes {
+    return _attributes;
   }
 }
--- a/lib/src/sdk/common/instrumentation_scope.dart
+++ b/lib/src/sdk/common/instrumentation_scope.dart
@@ -1,0 +1,19 @@
+class InstrumentationScope {
+  final String _name;
+  final String _version;
+  final String _schemaUrl;
+
+  InstrumentationScope(this._name, this._version, this._schemaUrl);
+
+  String get name {
+    return _name;
+  }
+
+  String get version {
+    return _version;
+  }
+
+  String get schemaUrl {
+    return _schemaUrl;
+  }
+}
--- a/lib/src/sdk/metrics/meter_provider.dart
+++ b/lib/src/sdk/metrics/meter_provider.dart
@@ -2,16 +2,23 @@ import 'package:opentelemetry/api.dart' as api;
 import 'package:opentelemetry/src/experimental_api.dart' as api;
 import 'package:opentelemetry/src/experimental_sdk.dart' as sdk;
 import 'package:logging/logging.dart';
-import 'package:opentelemetry/src/api/metrics/meter_key.dart';
+import 'package:opentelemetry/src/sdk/common/instrumentation_scope.dart';
+import 'package:opentelemetry/src/sdk/metrics/state/meter_provider_shared_state.dart';
 
 const invalidMeterNameMessage = 'Invalid Meter Name';
 
 class MeterProvider implements api.MeterProvider {
-  final _meters = <MeterKey, api.Meter>{};
   final _logger = Logger('opentelemetry.sdk.metrics.meterprovider');
+  final _shutdown = false;
+  final MeterProviderSharedState _sharedState;
+
+  sdk.Resource get resource => _sharedState.resource;
+
+  MeterProvider({sdk.Resource resource})
+      : _sharedState = MeterProviderSharedState(resource);
 
   @override
-  sdk.Meter get(String name,
+  api.Meter get(String name,
       {String version = '',
       String schemaUrl = '',
       List<api.Attribute> attributes = const []}) {
@@ -19,11 +26,15 @@ class MeterProvider implements api.MeterProvider {
       name = '';
       _logger.warning(invalidMeterNameMessage, '', StackTrace.current);
     }
-    version ??= '';
-    schemaUrl ??= '';
-    attributes ??= const [];
-    final key = MeterKey(name, version, schemaUrl, attributes);
 
-    return _meters.putIfAbsent(key, () => sdk.Meter());
+    if (_shutdown) {
+      _logger.warning('A shutdown MeterProvider cannot provide a Meter', '',
+          StackTrace.current);
+      return api.NoopMeter();
+    }
+
+    return _sharedState
+        .getMeterSharedState(InstrumentationScope(name, version, schemaUrl))
+        .meter;
   }
 }
--- a/lib/src/sdk/metrics/meter_provider.dart
+++ b/lib/src/sdk/metrics/meter_provider.dart
@@ -34,7 +34,7 @@ class MeterProvider implements api.MeterProvider {
     }
 
     return _sharedState
-        .getMeterSharedState(InstrumentationScope(name, version, schemaUrl))
+        .getMeterSharedState(InstrumentationScope(name, version, schemaUrl, attributes))
         .meter;
   }
 }
--- a/lib/src/sdk/metrics/state/meter_provider_shared_state.dart
+++ b/lib/src/sdk/metrics/state/meter_provider_shared_state.dart
@@ -4,8 +4,13 @@ import 'package:opentelemetry/src/sdk/metrics/state/meter_shared_state.dart';
 import 'package:quiver/core.dart';
 
 int instrumentationScopeId(InstrumentationScope instrumentationScope) {
-  return hash4(instrumentationScope.name, instrumentationScope.version,
-      instrumentationScope.schemaUrl, instrumentationScope.attributes);
+  return hash4(
+      instrumentationScope.name,
+      instrumentationScope.version,
+      instrumentationScope.schemaUrl,
+      instrumentationScope.attributes.isEmpty
+          ? const []
+          : instrumentationScope.attributes);
 }
 
 class MeterProviderSharedState {
--- a/lib/src/sdk/metrics/state/meter_provider_shared_state.dart
+++ b/lib/src/sdk/metrics/state/meter_provider_shared_state.dart
@@ -4,28 +4,23 @@ import 'package:opentelemetry/src/sdk/metrics/state/meter_shared_state.dart';
 import 'package:quiver/core.dart';
 
 int instrumentationScopeId(InstrumentationScope instrumentationScope) {
-  return hash4(
-      instrumentationScope.name,
-      instrumentationScope.version,
-      instrumentationScope.schemaUrl,
-      instrumentationScope.attributes.isEmpty
-          ? const []
-          : instrumentationScope.attributes);
+  return hash3(instrumentationScope.name, instrumentationScope.version,
+      instrumentationScope.schemaUrl);
 }
 
 class MeterProviderSharedState {
   Resource resource;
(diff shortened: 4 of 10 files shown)


aviary-wf:
## Security Insights

## Action Items
---

semveraudit-wf:
### Public API Changes

 No changes to the public API found for commit e9f49f0bcc3ee3b75007b532fff5a4ddcb44ed9b

Showing results for e9f49f0bcc3ee3b75007b532fff5a4ddcb44ed9b
[Browse public API](https://semver-audit.wk-dev.wdesk.org/Workiva/opentelemetry-dart/e9f49f0bcc3ee3b75007b532fff5a4ddcb44ed9b).

lib/src/sdk/metrics/state/meter_provider_shared_state.dart
```diff
@@ -0,0 +1,27 @@
+import 'package:opentelemetry/sdk.dart';
+import 'package:opentelemetry/src/sdk/common/instrumentation_scope.dart';
+import 'package:opentelemetry/src/sdk/metrics/state/meter_shared_state.dart';
+import 'package:quiver/core.dart';
+
+int instrumentationScopeId(InstrumentationScope instrumentationScope) {
+  return hash4(instrumentationScope.name, instrumentationScope.version,
+      instrumentationScope.schemaUrl, instrumentationScope.attributes);
```
blakeroberts-wk, line 8: Will this hash the address of instrumentationScope.attributes? If so, two distinct instantiations of instrumentation scope which have the same name/version/schemaUrl and empty attributes will still have two different instrumentationScopeIds. Is this intended?

dustinlessard-wf: It's not intended for InstrumentationScopeIds to be different when the name, version, schemaUrl and attributes are the same.  I've tested this scenario and found that when attributes are an empty list, the resulting hash4 is the same: [https://dartpad.dev/?id=5ea70f6e273105729a1f76938005bd1d](https://dartpad.dev/?id=5ea70f6e273105729a1f76938005bd1d)

blakeroberts-wk: That's because it's the same attributes variable. Make two attributes variables, each empty list, and the hashes will be different

dustinlessard-wf: I see now.  This is not intended behavior.  Reworking.

lib/src/experimental_sdk.dart
```diff
@@ -1,3 +1,7 @@
 export 'sdk/metrics/counter.dart' show Counter;
 export 'sdk/metrics/meter_provider.dart' show MeterProvider;
 export 'sdk/metrics/meter.dart' show Meter;
+export 'sdk/metrics/state/meter_shared_state.dart' show MeterSharedState;
+export 'sdk/metrics/state/meter_provider_shared_state.dart'
+    show MeterProviderSharedState;
+export 'sdk/resource/resource.dart' show Resource;
```
blakeroberts-wk, line 7: Should we mark all classes shown in this file (and in the experimental_api file) as `@experimental`?

dustinlessard-wf: I'm thinking that may not be necessary as the intended entry point for accessing these classes is via `experimental_sdk.dart` and `experimental_api.dart`.
It's true that one could access the code by referencing the `src` directory directly, but that's generally considered a bad practice.

blakeroberts-wk: 👍 do IDEs and other build tools recognize and report on usage of classes contained within a file prefixed with experimental?

lib/src/sdk/common/instrumentation_scope.dart
```diff
@@ -0,0 +1,27 @@
+import 'package:opentelemetry/api.dart' as api;
+
+class InstrumentationScope {
```
blakeroberts-wk, line 3: Is this just a newer https://github.com/Workiva/opentelemetry-dart/blob/master/lib/src/sdk/instrumentation_library.dart#L7? How should we handle the old one? Deprecate and update usage?

dustinlessard-wf: Definitely a similar concept, but they're not always the same thing.  [https://opentelemetry.io/docs/reference/specification/glossary/#instrumentation-scope](https://opentelemetry.io/docs/reference/specification/glossary/#instrumentation-scope) has a pretty ok description of how the two coexist.

lib/src/experimental_sdk.dart
dustinlessard-wf: I don't know of any IDE or build tools that key of classes contained within a file prefixed with experimental.  That's really for the engineer brave enough to use the code.  I tested out the 'Experimental' annotation on a function
`  @Experimental()
  void experiment() {}`
 And with a usage of the function ,neither VsCode nor Dart-Analyze gave any indication that I was using an experimental function.
 
Screen Shot 2022-11-11 at 5 18 07 PM

dustinlessard-wf: Screen Shot 2022-11-11 at 5 20 20 PM

lib/src/sdk/common/instrumentation_scope.dart
blakeroberts-wk: I think the concept of instrumentation library is just a concept and a concrete implementation is not needed. I think the concrete implementation is replaced by instrumentation scope:

- https://github.com/open-telemetry/opentelemetry-go/blob/308d0362e6c56cbbd23fa0fda14df5dbd70a5f51/CHANGELOG.md#deprecated
- https://github.com/open-telemetry/opentelemetry-go/blob/308d0362e6c56cbbd23fa0fda14df5dbd70a5f51/CHANGELOG.md#changed-10
- https://github.com/open-telemetry/opentelemetry-go/blob/575e1bb27025c73fd76f1e6b9dc2727b85867fdc/sdk/trace/snapshot.go#L105-L115

lib/src/experimental_sdk.dart
blakeroberts-wk: That might be a bug? This suggests that the analyzer should provide feedback:
https://pub.dev/documentation/meta/latest/meta/experimental-constant.html

dustinlessard-wf: @evanweible-wf , are you familiar with the 'experimental' annotation and whether the analyzer or IDE provide hints when it is in play?

lib/src/sdk/common/instrumentation_scope.dart
dustinlessard-wf: In light of [https://github.com/open-telemetry/opentelemetry-js/pull/2959](https://github.com/open-telemetry/opentelemetry-js/pull/2959) , I suggest we accept this for now and circle back to it once the metrics sdk is closer to non-experimental

lib/src/experimental_sdk.dart
evanweible-wf: Huh, sadly it looks like warning about experimental usages has not been implemented by the analyzer yet: https://github.com/dart-lang/sdk/issues/34258

But, looking at the current state of this branch, you've also not made these experimental entrypoints public, meaning that users would have to import from `src/` to get them. It is a well-understood convention in Dart that things within `src/` are internal and should not be imported (there's even a [lint](https://dart.dev/tools/linter-rules#implementation_imports) for it), so it probably accomplishes what you'd be going for with `@experimental`, just not as explicitly.

That being said, it wouldn't hurt to include the `@experimental` annotation and it might help make your intentions extra clear. If you do that, you'd put it at the library level in these files, like so:

```dart
@experimental
library experimental_sdk;

import 'package:meta/meta.dart';

...
```

lib/src/sdk/metrics/meter_provider.dart
```diff
@@ -2,28 +2,39 @@ import 'package:opentelemetry/api.dart' as api;
...
 
 const invalidMeterNameMessage = 'Invalid Meter Name';
 
 class MeterProvider implements api.MeterProvider {
-  final _meters = <MeterKey, api.Meter>{};
   final _logger = Logger('opentelemetry.sdk.metrics.meterprovider');
+  final _shutdown = false;
+  final MeterProviderSharedState _sharedState;
+
+  sdk.Resource get resource => _sharedState.resource;
```
evanweible-wf, line 15: Similar to the change below for `get()`, should this return be typed `api.Resource`?

lib/src/sdk/metrics/state/meter_provider_shared_state.dart
```diff
@@ -0,0 +1,32 @@
...
+import 'package:quiver/core.dart';
+
+int instrumentationScopeId(InstrumentationScope instrumentationScope) {
+  return hash4(
+      instrumentationScope.name,
+      instrumentationScope.version,
+      instrumentationScope.schemaUrl,
+      instrumentationScope.attributes.isEmpty
+          ? const []
+          : instrumentationScope.attributes);
```
evanweible-wf, line 13: It wasn't clear to me at first that this use of `const []` was necessary, but I see from Blake's original comment that it is required to ensure that empty attribute lists are treated as identical. So at the very least, I think this is worth a code comment explaining that.

However, the [glossary link you referenced above](https://opentelemetry.io/docs/reference/specification/glossary/#instrumentation-scope) seems to suggest that a scope is typically defined by a name, or a name + version, and may also include zero or more additional attributes that provide additional information about the scope. But it's not clear to me whether those additional attributes should be included when determining uniqueness. Is there an example from another SDK that we can follow?

lib/src/sdk/metrics/meter_provider.dart
```diff
@@ -2,28 +2,39 @@ import 'package:opentelemetry/api.dart' as api;
...
-    schemaUrl ??= '';
-    attributes ??= const [];
-    final key = MeterKey(name, version, schemaUrl, attributes);
 
-    return _meters.putIfAbsent(key, () => sdk.Meter());
+    if (_shutdown) {
+      _logger.warning('A shutdown MeterProvider cannot provide a Meter', '',
+          StackTrace.current);
+      return api.NoopMeter();
+    }
```
evanweible-wf, line 34: What does the lifecycle of a provider look like? Currently doesn't look like `_shutdown` is ever set to true, will that be added later?

lib/src/sdk/metrics/state/meter_provider_shared_state.dart
```diff
@@ -0,0 +1,32 @@
...
+}
+
+class MeterProviderSharedState {
+  Resource resource;
+  Map<int, MeterSharedState> meterSharedStates = {};
```
evanweible-wf, line 18: Should this be private?

lib/src/sdk/metrics/meter_provider.dart
evanweible-wf: Hmm.. I guess the spec only talks about an SDK Resource. Seems a bit odd, though - what do other OTel libs do?

lib/src/sdk/metrics/state/meter_shared_state.dart
```diff
@@ -0,0 +1,15 @@
...
+  // ignore: unused_field
+  final sdk.MeterProviderSharedState _meterProviderSharedState;
+  // ignore: unused_field
+  final InstrumentationScope _instrumentationScope;
+  sdk.Meter meter;
+
+  MeterSharedState(this._meterProviderSharedState, this._instrumentationScope) {
+    meter = sdk.Meter(this);
+  }
+}
```
evanweible-wf, line 15: I'm struggling to understand why these shared state classes are necessary. I see that the spec requires that you should be able to associate a MeterProvider with a Resource upon construction, and that any meter created from that provider MUST be associated with the same Resource. Are there other similar associations at the provider level that we'll need to add in the future? What is the value of abstracting that association into a class like `MeterProviderSharedState`?

I'm similarly not following why `MeterSharedState` is needed. The provider should be able to pass on its association with a Resource when creating Meters, and the association between the Meter and an InstrumentationScope is known when constructing the Meter, so could it be stored on the Meter directly?

The only other missing piece is ensuring that the InstrumentationScope uniquely identifies Meters so that the provider can return an existing Meter when applicable, but that map of scopes to meters could be stored directly on the provider instead of in a shared state class.

lib/src/sdk/common/instrumentation_scope.dart
dustinlessard-wf: https://github.com/Workiva/opentelemetry-dart/issues/98

lib/src/sdk/metrics/meter_provider.dart
dustinlessard-wf: Java includes the Resource class only within the SDK.  JS is super murky to me on what is SDK and what is API, but their `resource` depends on the "SDK" which implies to me that it is also part of the SDK.

[The API is intended for library code to use whereas the SDK is intended for application code to use.](https://opentelemetry.io/docs/concepts/instrumenting-library/#opentelemetry-api)  Since the API is really just a collection of abstract classes/interfaces + noop implementations, it seems like the intention is for library code to not create MetricProviders of their own.

lib/src/sdk/metrics/meter_provider.dart
dustinlessard-wf: Yes, the code here is a partial implementation intended to build towards being able to create a `Counter`.  The shutdown functionality is a future concern.

lib/src/sdk/metrics/state/meter_shared_state.dart
dustinlessard-wf: I've been working on a much larger implementation that allows synchronous counters to be created.  I've included these two classes in this pr not because they are necessary for the inclusion of the resource functionality, but because we are essentially duplicating the OpenTelemetryJS implementation in Dart and I'm trying to get code pr'd toward that larger implementation incrementally.  If we are to follow the design from OTELJS, then we would say that the MeterProviderSharedState is critical for the [views api](https://github.com/open-telemetry/opentelemetry-js/pull/2625), and the MeterSharedState required for a future inclusion of Metric Storages.

This is all wip, but my concern with pr'ing it all at once is that it would make the pr a monster, and I'd imagine it will be far easier to miss issues if I go in that direction.

Thoughts?

evanweible-wf: Gotcha. I definitely agree with keeping the PRs small, no need to balloon this one just to see the end goal for the shared state classes. As long as they're kept as implementation details, we can always adjust or at least have a more applicable conversation on a future PR. That being said, it looks like the two shared state classes are currently exported from the `experimental_sdk.dart` entrypoint, but I don't think they need to be. Consumers shouldn't need to know about them, right?

dustinlessard-wf: True, consumers do not need to know about them. I'll remove.

lib/src/sdk/metrics/state/meter_provider_shared_state.dart
dustinlessard-wf: That's a good point.  I've asked folks in the OTEL Specification space their thoughts on this.  Given this:

>  Since the scope is a build-time concept the attributes of the scope cannot change at runtime.

I don't see a compelling reason to include the attributes to identify a given scope.  I'll change that.

lib/src/sdk/metrics/state/meter_provider_shared_state.dart
dustinlessard-wf: fixed

lib/src/sdk/metrics/state/meter_shared_state.dart
dustinlessard-wf: Fixed

evanweible-wf (approved):

dustinlessard-wf:
QA +1
@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #109, opened 2023-05-19
Move W3CTraceContextPropagator to API package

michaelyeager-wf:
## Which problem is this PR solving?

The propagator for W3C-format traces is present in the SDK portion of this package which does not conform to [the OpenTelemetry specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/api-propagators.md#propagators-distribution). 

## Short description of the change

This PR moves `W3CTracePropagator` to the API portion of this package to conform to the OpenTelemetry specification.

## How Has This Been Tested?

Unit tests demonstrate that these changes operate correctly.

## Checklist:

- [ ] Unit tests have been added
- [ ] Documentation has been updated

PR #118, opened 2023-09-13
O11Y-3138: Improve Adherence to Specification

michaelyeager-wf:

This library does not closely adhere to the OpenTelemetry Specification.  To improve adherence to the specification and lay the ground-work for null-safety, this PR makes the following changes:

* Fields in `TracerProviderBase` are now protected rather than private.  
* `WebTracerProvider` now uses fields on its super class, where possible.
* Constructor for `SpanLimits` is now constant.
* Added `NoopTextMapPropagator`, which `globalTextMapPropagator` now uses by default.
* Added `NoopTracerProvider`, which `globalTracerProvider` now uses by default.
* Began the process of migrating `Span.setStatus` to `Span.setStatusCode`.  The latter includes an updated function signature which uses positional optional arguments instead of named optional arguments.
* Added deprecation messages for planned future breaking changes.
* Updated references to SDK `SpanContext` and `TraceState` in unrelated unit tests.  These classes were moved to API in #109 and should no longer be used.

## How Has This Been Tested?

* Confirmed passing unit tests
* Updated a test application from a previous version.  Confirmed that changes are non-breaking with no further updates.

### DIFF
--- a/lib/src/api/trace/nonrecording_span.dart
+++ b/lib/src/api/trace/nonrecording_span.dart
@@ -13,7 +13,7 @@ import '../../../api.dart' as api;
 /// This class should not be exposed to consumers and is used internally to wrap
 /// [api.SpanContext] being injected or extracted for external calls.
 @Deprecated(
-    'This class will stop being exported in a future release.  Please migrate to use of Span instead')
+    'This class will stop being exported in a future release.  Please use [api.Span] instead.')
 class NonRecordingSpan implements api.Span {
   final api.SpanStatus _status = api.SpanStatus()..code = api.StatusCode.ok;
   final api.SpanContext _spanContext;
@@ -45,7 +45,12 @@ class NonRecordingSpan implements api.Span {
   api.SpanId get parentSpanId => api.SpanId.invalid();
 
   @override
+  @Deprecated(
+      'This method will be removed in a future release.  Use [NonRecordingSpan.setStatusCode] instead.')
   void setStatus(api.StatusCode status, {String description}) {}
+
+  @override
+  void setStatusCode(api.StatusCode status, [String description]) {}
 
   @override
   api.SpanContext get spanContext => _spanContext;
--- a/lib/src/api/trace/nonrecording_span.dart
+++ b/lib/src/api/trace/nonrecording_span.dart
@@ -46,11 +46,8 @@ class NonRecordingSpan implements api.Span {
 
   @override
   @Deprecated(
-      'This method will be removed in a future release.  Use [NonRecordingSpan.setStatusCode] instead.')
+      'This method will be updated to use positional optional parameters in a future release.')
   void setStatus(api.StatusCode status, {String description}) {}
-
-  @override
-  void setStatusCode(api.StatusCode status, [String description]) {}
 
   @override
   api.SpanContext get spanContext => _spanContext;
--- a/lib/src/api/trace/nonrecording_span.dart
+++ b/lib/src/api/trace/nonrecording_span.dart
@@ -13,7 +13,7 @@ import '../../../api.dart' as api;
 /// This class should not be exposed to consumers and is used internally to wrap
 /// [api.SpanContext] being injected or extracted for external calls.
 @Deprecated(
-    'This class will stop being exported in a future release.  Please use [api.Span] instead.')
+    'This class will stop being exported in v0.18.0.  Please use [api.Span] instead.')
 class NonRecordingSpan implements api.Span {
   final api.SpanStatus _status = api.SpanStatus()..code = api.StatusCode.ok;
   final api.SpanContext _spanContext;
@@ -46,7 +46,7 @@ class NonRecordingSpan implements api.Span {
 
   @override
   @Deprecated(
-      'This method will be updated to use positional optional parameters in a future release.')
+      'This method will be updated to use positional optional parameters in v0.18.0.')
   void setStatus(api.StatusCode status, {String description}) {}
 
   @override
--- a/lib/src/api/trace/nonrecording_span.dart
+++ b/lib/src/api/trace/nonrecording_span.dart
@@ -12,6 +12,8 @@ import '../../../api.dart' as api;
 ///
 /// This class should not be exposed to consumers and is used internally to wrap
 /// [api.SpanContext] being injected or extracted for external calls.
+@Deprecated(
+    'This class will stop being exported in a future release.  Please migrate to use of Span instead')
 class NonRecordingSpan implements api.Span {
   final api.SpanStatus _status = api.SpanStatus()..code = api.StatusCode.ok;
   final api.SpanContext _spanContext;
--- a/lib/src/api/trace/span.dart
+++ b/lib/src/api/trace/span.dart
@@ -28,6 +28,8 @@ enum SpanKind {
 /// function calls to sub-components. A trace has a single, top-level "root"
 /// span that in turn may haze zero or more child Spans, which in turn may have
 /// children.
+///
+/// Warning: methods may be added to this interface in minor releases.
 abstract class Span {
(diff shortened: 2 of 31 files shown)


aviary3-wk:
## Security Insights

## Action Items
---

semveraudit-wf:
### Public API Changes

```diff
@@ opentelemetry/api.dart <-- package:opentelemetry/src/api/trace/span.dart#L34 @@
abstract class Span
+     void setStatusCode(StatusCode status, [String description])
//    Adding abstract members breaks all subclasses.

```

```diff
@@ opentelemetry/api.dart <-- package:opentelemetry/src/api/trace/nonrecording_span.dart#L18 @@
class NonRecordingSpan implements Span
+     void setStatusCode(StatusCode status, [String description])
//    Adding a method is a minor change.

```

```diff
@@ opentelemetry/sdk.dart <-- package:opentelemetry/src/sdk/trace/tracer_provider.dart#L12 @@
class TracerProviderBase implements TracerProvider
+     IdGenerator get idGenerator
//    Adding a field is a minor change.

+     Map<String, Tracer> get tracers
//    Adding a field is a minor change.

+     SpanLimits get spanLimits
//    Adding a field is a minor change.

+     Resource get resource
//    Adding a field is a minor change.

+     Sampler get sampler
//    Adding a field is a minor change.

+     List<SpanProcessor> get processors
//    Adding a field is a minor change.

```

```diff
@@ opentelemetry/sdk.dart <-- package:opentelemetry/src/sdk/trace/span.dart#L14 @@
class Span implements Span
+     void setStatusCode(StatusCode status, [String description])
//    Adding a method is a minor change.

```

```diff
@@ opentelemetry/web_sdk.dart <-- package:opentelemetry/src/sdk/platforms/web/trace/web_tracer_provider.dart#L21 @@
class WebTracerProvider extends TracerProviderBase

      From package:opentelemetry/src/sdk/trace/tracer_provider.dart
+     Sampler get sampler
//    Adding a field is a minor change.
+     List<SpanProcessor> get processors
//    Adding a field is a minor change.
+     Map<String, Tracer> get tracers
//    Adding a field is a minor change.
+     SpanLimits get spanLimits
//    Adding a field is a minor change.
+     Resource get resource
//    Adding a field is a minor change.
+     IdGenerator get idGenerator
//    Adding a field is a minor change.

```

Showing results for 531641b77bb19636376251d8542cfbdc349cd22d
[Browse public API](https://semver-audit.wk-dev.wdesk.org/Workiva/opentelemetry-dart/531641b77bb19636376251d8542cfbdc349cd22d).

lib/src/sdk/trace/span.dart
```diff
@@ -89,6 +95,22 @@ class Span implements api.Span {
     }
   }
 
+  @override
+  void setStatusCode(api.StatusCode status, [String description]) {
```
blakeroberts-wk, line 99: nit (arbitrary line): could replace the body of the deprecated method with a call to this new non-deprecated method signifying to devs that the functionality is identical

lib/src/api/trace/span.dart
```diff
@@ -62,8 +62,19 @@ abstract class Span {
...
   void setStatus(api.StatusCode status, {String description});
 
+  /// Sets the status to the [Span].
+  ///
+  /// If used, this will override the default [Span] status. Default status code
+  /// is [api.StatusCode.unset].
+  ///
+  /// Only the value of the last call will be recorded, and implementations are
+  /// free to ignore previous calls.
+  void setStatusCode(api.StatusCode status, [String description]);
```
blakeroberts-wk, line 76: Rather than moving from `setStatus` to `setStatusCode` to switch from optional named args to optional positional args, I suggest we:
- keep `setStatus` with its current signature for this release
- do not add `setStatusCode`
- update the deprecation warning to callout that the next minor release of this library will switch from optional named args to optional positional args
- when the next minor version is cut, include the breaking change as mentioned in the deprecation warning
 
Because this is the only breaking change in the PR, this'll allow us to release this PR under a patch version update.

Of course, this means that there won't exist an intermediate release that allows consumers to update the version of the library without updating their code (for the next minor release which includes the breaking change). I think this pain is permissible because this library is pre-1.0 and so development should prioritize longterm API health over short term version upgrades (when the two conflict).

lib/src/api/trace/span.dart
```diff
@@ -49,7 +49,7 @@ abstract class Span {
   String name;
```
blakeroberts-wk, line 49: arbitrary line: In at least otel-go, the public api interfaces are marked with warnings that methods may be added in minor releases (post 1.0). I think we should add this warning as well.

https://github.com/open-telemetry/opentelemetry-go/blob/main/trace/trace.go#L341

lib/src/api/trace/span.dart
michaelyeager-wf: I was thinking about this yesterday and I agree that this approach is better.  I have an update staged with breaking changes already, so adding another breaking change won't affect things much at all.

lib/src/api/trace/nonrecording_span.dart
```diff
@@ -43,6 +45,8 @@ class NonRecordingSpan implements api.Span {
   api.SpanId get parentSpanId => api.SpanId.invalid();
 
   @override
+  @Deprecated(
+      'This method will be updated to use positional optional parameters in a future release.')
```
blakeroberts-wk, line 49:
```suggestion
      'This method will be updated to use positional optional parameters in v0.17.0.')
```

lib/src/api/trace/span.dart
```diff
@@ -49,7 +51,7 @@ abstract class Span {
...
 
   /// Whether this Span is recording information like events with the
-  /// addEvent operation, status with setStatus, etc.
+  /// addEvent operation, status with setStatusCode, etc.
```
blakeroberts-wk, line 54:
```suggestion
  /// addEvent operation, status with setStatus, etc.
```

lib/src/api/trace/span.dart
```diff
@@ -28,6 +28,8 @@ enum SpanKind {
 /// function calls to sub-components. A trace has a single, top-level "root"
 /// span that in turn may haze zero or more child Spans, which in turn may have
 /// children.
+///
+/// Warning: methods may be added to this interface in minor releases.
```
blakeroberts-wk, line 32: nit: can we add this comment to api.Tracer and api.TracerProvider in this PR?

lib/src/sdk/trace/span.dart
```diff
@@ -74,6 +78,8 @@ class Span implements api.Span {
   }
 
   @override
+  @Deprecated(
+      'This method will be updated to use positional optional parameters in a future release.')
```
blakeroberts-wk, line 82:
```suggestion
      'This method will be updated to use positional optional parameters in v0.17.0.')
```

michaelyeager-wf: That'll be v0.18.0 with what I have staged, provided nothing else merges in the mean time.  I can make this change, but I'd prefer not to give exact version numbers for that reason.

lib/src/api/trace/nonrecording_span.dart
michaelyeager-wf: That'll be v0.18.0 with what I have staged, provided nothing else merges in the mean time.  I can make this change, but I'd prefer not to give exact version numbers for that reason.

blakeroberts-wk (approved):

blakeroberts-wk (approved):

dustinlessard-wf:
🎉 🙌 ✊ 🏆

dustinlessard-wf (approved):

robbecker-wf:
+10 @Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #125, opened 2023-09-29
O11Y-1786: Implement Read-Only Spans

michaelyeager-wf:

This PR implements read-only Spans in order to prevent modification during processing and export.

* API:  Added `ReadableSpan`, which contains the read-able portions formerly contained in `Span`.
* API:  `Span` now extends `ReadableSpan` and adds write-able portions.
* SDK:  Added `ReadableSpan` which extends API's `ReadableSpan` to add readable, non-consumer-facing properties like links and attributes.
* SDK:  `Span` now implements API's `Span` and SDK's `ReadableSpan`.
* SDK:  Processors and Exporters now expect `ReadableSpan`s.

## How Has This Been Tested?

* Confirmed passing unit tests
* Updated a test application from a previous version. Confirmed that changes operate properly with no further updates.

### DIFF
--- a/lib/src/api/trace/span.dart
+++ b/lib/src/api/trace/span.dart
@@ -38,24 +38,11 @@ abstract class Span {
   /// span ends.
   api.SpanContext get spanContext;
 
-  /// Get the time when the span was closed, or null if still open.
-  Int64 get endTime;
-
-  /// Get the time when the span was started.
-  Int64 get startTime;
-
   /// The parent span id.
   api.SpanId get parentSpanId;
 
-  /// The name of the span.
-  String name;
-
-  /// Whether this Span is recording information like events with the
-  /// addEvent operation, status with setStatus, etc.
-  bool get isRecording;
-
-  /// The kind of the span.
-  SpanKind get kind;
+  /// Sets the name of the [Span].
+  void setName(String name);
 
   /// Sets the status to the [Span].
   ///
@@ -68,17 +55,11 @@ abstract class Span {
       'This method will be updated to use positional optional parameters in v0.17.0.')
   void setStatus(api.StatusCode status, {String description});
 
-  /// Retrieve the status of the [Span].
-  api.SpanStatus get status;
-
   /// set single attribute
   void setAttribute(api.Attribute attribute);
 
   /// set multiple attributes
   void setAttributes(List<api.Attribute> attributes);
-
-  /// Retrieve the instrumentation library on this span.
-  api.InstrumentationLibrary get instrumentationLibrary;
 
   /// Record metadata about an event occurring during this span.
   void addEvent(String name, Int64 timestamp, {List<api.Attribute> attributes});
--- a/lib/src/sdk/platforms/web/trace/web_tracer_provider.dart
+++ b/lib/src/sdk/platforms/web/trace/web_tracer_provider.dart
@@ -36,7 +36,7 @@ class WebTracerProvider extends sdk.TracerProviderBase {
             spanLimits: spanLimits ?? sdk.SpanLimits());
 
   @override
-  Tracer getTracer(String name, {String version = ''}) {
+  api.Tracer getTracer(String name, {String version = ''}) {
     return tracers.putIfAbsent(
         '$name@$version',
         () => Tracer(processors, resource, sampler, _timeProvider, idGenerator,
--- a/lib/src/sdk/platforms/web/trace/web_tracer_provider.dart
+++ b/lib/src/sdk/platforms/web/trace/web_tracer_provider.dart
@@ -19,7 +19,7 @@ class WebTracerProvider extends sdk.TracerProviderBase {
   final sdk.TimeProvider _timeProvider;
 
   WebTracerProvider(
-      {List<api.SpanProcessor> processors,
+      {List<sdk.SpanProcessor> processors,
       sdk.Resource resource,
       sdk.Sampler sampler,
       sdk.TimeProvider timeProvider,
@@ -36,7 +36,7 @@ class WebTracerProvider extends sdk.TracerProviderBase {
             spanLimits: spanLimits ?? sdk.SpanLimits());
 
   @override
-  api.Tracer getTracer(String name, {String version = ''}) {
+  Tracer getTracer(String name, {String version = ''}) {
     return tracers.putIfAbsent(
         '$name@$version',
         () => Tracer(processors, resource, sampler, _timeProvider, idGenerator,
--- a/lib/src/sdk/trace/span.dart
+++ b/lib/src/sdk/trace/span.dart
@@ -10,12 +10,12 @@ import '../common/attributes.dart';
 /// A representation of a single operation within a trace.
 @Deprecated(
     'This class will stop being exported and be marked protected in v0.17.0.  Consumers should use [api.Span] instead.')
-class Span implements api.Span {
+class Span implements sdk.ReadWriteSpan {
   final api.SpanContext _spanContext;
   final api.SpanId _parentSpanId;
   final api.SpanKind _kind;
   final api.SpanStatus _status = api.SpanStatus();
-  final List<api.SpanProcessor> _processors;
+  final List<sdk.SpanProcessor> _processors;
   final List<api.SpanLink> _links;
   final sdk.TimeProvider _timeProvider;
   final sdk.Resource _resource;
@@ -23,19 +23,24 @@ class Span implements api.Span {
   final api.InstrumentationLibrary _instrumentationLibrary;
   final Int64 _startTime;
   final Attributes _attributes = Attributes.empty();
+  String _name;
   Int64 _endTime;
   int _droppedSpanAttributes = 0;
 
   @override
-  String name;
+  void setName(String name) {
+    _name = name;
+  }
 
   @override
+  String get name => _name;
+
   bool get isRecording => _endTime == null;
 
   /// Construct a [Span].
   @Deprecated(
       'This constructor will be marked protected in v0.17.0.  Consumers should use [api.Span] instead.')
-  Span(this.name, this._spanContext, this._parentSpanId, this._processors,
+  Span(this._name, this._spanContext, this._parentSpanId, this._processors,
       this._timeProvider, this._resource, this._instrumentationLibrary,
       {api.SpanKind kind,
       List<api.Attribute> attributes,
@@ -98,6 +103,7 @@ class Span implements api.Span {
   @override
   api.SpanStatus get status => _status;
 
+  @override
   sdk.Resource get resource => _resource;
 
   @override
@@ -223,8 +229,10 @@ class Span implements api.Span {
     return spanLink;
   }
 
-  List<api.SpanLink> get links => _links;
+  @override
+  List<api.SpanLink> get links => List.unmodifiable(_links);
 
+  @override
   Attributes get attributes => _attributes;
 
   //Truncate just strings which length is longer than configuration.
--- a/lib/src/sdk/trace/span_processors/span_processor.dart
+++ b/lib/src/sdk/trace/span_processors/span_processor.dart
@@ -1,14 +1,13 @@
 // Copyright 2021-2022 Workiva.
 // Licensed under the Apache License, Version 2.0. Please see https://github.com/Workiva/opentelemetry-dart/blob/master/LICENSE for more information
 
-import '../../../api.dart' as api;
+import '../../../../api.dart' as api;
+import '../../../../sdk.dart' as sdk;
 
-@Deprecated(
-    'This class will be moved to the SDK package in v0.18.0.  Use [SpanExporter] from SDK instead.')
 abstract class SpanProcessor {
-  void onStart(api.Span span, api.Context parentContext);
+  void onStart(sdk.ReadWriteSpan span, api.Context parentContext);
 
-  void onEnd(api.Span span);
+  void onEnd(sdk.ReadOnlySpan span);
 
   void shutdown();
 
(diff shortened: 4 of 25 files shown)


aviary-wf:
## Security Insights

## Action Items
---

semveraudit-wf:
### Public API Changes

```diff
@@ opentelemetry/sdk.dart <-- package:opentelemetry/src/sdk/trace/exporters/collector_exporter.dart#L18 @@
class CollectorExporter implements SpanExporter
-     void export(List<Span> spans)
+     void export(List<ReadableSpan> spans)
//    `type` of `spans` has changed.
//    Changing a parameter signature is a major change.

```

```diff
@@ opentelemetry/sdk.dart <-- package:opentelemetry/src/sdk/trace/span_processors/batch_processor.dart#L14 @@
class BatchSpanProcessor implements SpanProcessor
-     void onEnd(Span span)
+     void onEnd(ReadableSpan span)
//    `type` of `span` has changed.
//    Changing a parameter signature is a major change.

```

```diff
@@ opentelemetry/api.dart <-- package:opentelemetry/src/api/exporters/span_exporter.dart#L9 @@
abstract class SpanExporter
-     void export(List<Span> spans)
+     void export(List<ReadableSpan> spans)
//    `type` of `spans` has changed.
//    Changing a parameter signature is a major change.

```

```diff
@@ opentelemetry/api.dart <-- package:opentelemetry/src/api/span_processors/span_processor.dart#L9 @@
abstract class SpanProcessor
-     void onEnd(Span span)
+     void onEnd(ReadableSpan span)
//    `type` of `span` has changed.
//    Changing a parameter signature is a major change.

```

```diff
@@ opentelemetry/sdk.dart <-- package:opentelemetry/src/sdk/trace/exporters/console_exporter.dart#L11 @@
class ConsoleExporter implements SpanExporter
-     void export(List<Span> spans)
+     void export(List<ReadableSpan> spans)
//    `type` of `spans` has changed.
//    Changing a parameter signature is a major change.

```
<details>
<summary>Click to see 1 more API Changes</summary><p>

---
```diff
@@ opentelemetry/sdk.dart <-- package:opentelemetry/src/sdk/trace/span_processors/simple_processor.dart#L9 @@
class SimpleSpanProcessor implements SpanProcessor
-     void onEnd(Span span)
+     void onEnd(ReadableSpan span)
//    `type` of `span` has changed.
//    Changing a parameter signature is a major change.

```
---

</p></details>

Showing results for 1658fe0b0dad010a1f315dd364cb54c43858eabb
[Browse public API](https://semver-audit.wk-dev.wdesk.org/Workiva/opentelemetry-dart/1658fe0b0dad010a1f315dd364cb54c43858eabb).

lib/src/sdk/trace/span.dart
```diff
@@ -223,8 +227,10 @@ class Span implements api.Span {
     return spanLink;
   }
 
-  List<api.SpanLink> get links => _links;
+  @override
+  List<api.SpanLink> get links => List.unmodifiable(_links);
```
michaelyeager-wf, line 231: Added this in because by the spec links aren't supposed to be able to be changed after being added.  This method allows us to still use a `List` for API purposes, but I believe it works by `throw`-ing on modification, which may not be cool.  We could also look at something like `UnmodifiableListView` instead or tackle this point separately.

lib/src/api/trace/readable_span.dart
```diff
@@ -0,0 +1,39 @@
+// Copyright 2021-2022 Workiva.
+// Licensed under the Apache License, Version 2.0. Please see https://github.com/Workiva/opentelemetry-dart/blob/master/LICENSE for more information
+
+import 'package:fixnum/fixnum.dart';
+
+import '../../../api.dart' as api;
+
+/// A representation of readable information about a single operation within a trace.
```
blakeroberts-wk, line 8: We should add the warning here about potential additions to the interface in minor releases

lib/src/api/trace/readable_span.dart
```diff
@@ -0,0 +1,39 @@
...
+abstract class ReadableSpan {
```
blakeroberts-wk, line 9: nit:
```suggestion
abstract class ReadOnlySpan {
```

lib/src/api/exporters/span_exporter.dart
```diff
@@ -1,10 +1,10 @@
 // Copyright 2021-2022 Workiva.
 // Licensed under the Apache License, Version 2.0. Please see https://github.com/Workiva/opentelemetry-dart/blob/master/LICENSE for more information
 
-import '../../../api.dart' as api;
+import '../trace/readable_span.dart';
 
 abstract class SpanExporter {
```
blakeroberts-wk, line 6: This should be moved to the sdk

lib/src/api/span_processors/span_processor.dart
```diff
@@ -1,12 +1,14 @@
 // Copyright 2021-2022 Workiva.
 // Licensed under the Apache License, Version 2.0. Please see https://github.com/Workiva/opentelemetry-dart/blob/master/LICENSE for more information
 
+import '../trace/readable_span.dart';
+
 import '../../../api.dart' as api;
 
 abstract class SpanProcessor {
```
blakeroberts-wk, line 8: This should be moved to the sdk

lib/src/api/trace/readable_span.dart
blakeroberts-wk, line 9: This should be moved to the sdk

lib/src/api/trace/span.dart
```diff
@@ -30,33 +31,14 @@ enum SpanKind {
...
-  /// Get the time when the span was started.
-  Int64 get startTime;
-
-  /// The parent span id.
-  api.SpanId get parentSpanId;
-
-  /// The name of the span.
+abstract class Span extends ReadableSpan {
+  @override
   String name;
```
blakeroberts-wk, line 36: We should avoid making the span name readable from the type `api.Span`

lib/src/api/trace/span.dart
```diff
@@ -30,33 +31,14 @@ enum SpanKind {
...
-  Int64 get endTime;
-
-  /// Get the time when the span was started.
-  Int64 get startTime;
-
-  /// The parent span id.
-  api.SpanId get parentSpanId;
-
-  /// The name of the span.
+abstract class Span extends ReadableSpan {
```
blakeroberts-wk, line 34: I don't think we want to extend here:
- `api.Span` should be write-only.
- The read-only span is intended for implementing exporters and processors only.
- There could be a read/write span type defined by the sdk. I suppose this would be `sdk.Span`, but another type could be used. I think ideally either of these types would be protected, but that may not be possible.

lib/src/api/trace/readable_span.dart
michaelyeager-wf: Doing this and keeping api.Span write-only leaves api.NonRecordingSpan unusable without further work.  I'll look for ways to mitigate that.

lib/src/api/trace/span.dart
michaelyeager-wf: >api.Span should be write-only

Implementing this would introduce limitations which aren't present currently without further work.  For example, reading SpanContext for Spans stored in api.Context, which is done when new Spans are created.  I'll look for ways to mitigate this.

lib/src/api/trace/span.dart
michaelyeager-wf: Dart heavily discourages setter-only fields, so we'll want to do something like `setName` instead.  I'll look at this more closely.

blakeroberts-wk: I think that makes sense. Other otel sdk's use `setName` so there's parity there

lib/src/api/trace/readable_span.dart
michaelyeager-wf: I'm still not sure about this.  We don't want to export it, to be sure, but I think we want to keep the readable, consumer-accessible parts of "Span" defined here so that both NonRecordingSpan (API) and Span (SDK) can keep in sync on the same base class.  My concern is what we'll run into trying to maintain the both of them without that.

blakeroberts-wk: The non recording span should be a minimal implementation if api.Span. Maintaining its implementation should be very low effort. Checkout what Go looks like:
- https://github.com/open-telemetry/opentelemetry-go/blob/main/trace/nonrecording.go#L20-L24
- https://github.com/open-telemetry/opentelemetry-go/blob/main/trace/noop.go#L57-L89

michaelyeager-wf: >The non recording span should be a minimal implementation if api.Span.
Sure, and it is currently.  But going forward, API's Span will only include the write-able portions of a "Span"'s API.  If the _readable_ portions of "Span" only exist in the SDK, how would we like to ensure that the readable portions of NonRecordingSpan and SDK's Span remain equivalent?  "Remember to change both of them" isn't a good approach, I don't think.

blakeroberts-wk: There shouldn't need to be readable parts of a non-recording span except for the span context which is apart of the api.Span interface

blakeroberts-wk: We may need to redefine a non-recording span type for both the api and sdk. The api would return this type only from a no-op tracer. The sdk would return its non-recording span if the span isn't sampled (by the sampling result determined by the tracer's provider's sampler).

The only time a span's content needs to be read is from a processor during onStart() or onEnd(). Both of these are only called with recording spans so non-recording spans don't need fields that can be read aside from the span context used for trace context propagation

lib/src/sdk/platforms/web/trace/web_tracer_provider.dart
```diff
@@ -36,7 +36,7 @@ class WebTracerProvider extends sdk.TracerProviderBase {
             spanLimits: spanLimits ?? sdk.SpanLimits());
 
   @override
-  api.Tracer getTracer(String name, {String version = ''}) {
+  Tracer getTracer(String name, {String version = ''}) {
```
blakeroberts-wk, line 39: I believe a tracer provider should implement an `api.TracerProvider` which has a method `getTracer()` that returns an `api.Tracer`

lib/src/sdk/trace/span.dart
blakeroberts-wk: I think this makes sense. Can links be added after a span has started? If so, we'll need an `addLink()` or similar. Additionally, I think we should add `List.unmodifiable()` to the attributes getter as well? Any attribute setting should be done via `setAttribute()` to enforce attribute limits

lib/src/sdk/trace/span_processors/span_processor.dart
```diff
@@ -0,0 +1,15 @@
+// Copyright 2021-2022 Workiva.
+// Licensed under the Apache License, Version 2.0. Please see https://github.com/Workiva/opentelemetry-dart/blob/master/LICENSE for more information
+
+import '../../../../api.dart' as api;
+import '../../../../sdk.dart' as sdk;
+
+abstract class SpanProcessor {
```
blakeroberts-wk, line 7: Not actionable, just thinking out loud: wondering how helpful it'd be to link to specification 🤔 

Again, not actionable here. But something like the following (or similar if there's a more appropriate syntax for dart doc comments):

```suggestion
/// Processes [api.Span]s on start and end.
///
/// * [Specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#span-processor)
abstract class SpanProcessor {
```

lib/src/sdk/trace/tracer.dart
```diff
@@ -25,7 +25,7 @@ class Tracer implements api.Tracer {
       : _spanLimits = spanLimits ?? sdk.SpanLimits();
 
   @override
-  api.Span startSpan(String name,
+  Span startSpan(String name,
```
blakeroberts-wk, line 28: Does dart allow implicit implementations? I didn't think that was supported and so since `api.Tracer` defines the `startSpan` method as returning `api.Span`, then the implementation here must also return `api.Span`. Is that not the case?

michaelyeager-wf: Since API's Span is an interface and SDK's Span implements it, SDK's Span _is_ an API Span.  Done this way here because I assumed we would need access to read-able portions of Span before processing/export within the SDK and also to avoid casting in the unit tests, which I wasn't excited about.

blakeroberts-wk: > Since API's Span is an interface and SDK's Span implements it, SDK's Span is an API Span.

This is implicit implementation. Which I didn't think Dart supported it. Like I'd assume that the dart analyzer would fail on this. And even if it doesn't, this probably means that consumers will end up with an `sdk.Span` which isn't a public/exported type and they'd have to cast it:

```dart
let span = tracer.startSpan("my-span") as api.Span;
```

michaelyeager-wf: I see your point.  I think we can stand a little casting in tests to prevent that.

lib/src/sdk/trace/span_processors/span_processor.dart
michaelyeager-wf: 🤔 Might be a bit of a pain to keep up-to-date, but it'd also be handy to not have to go searching for the spec.

lib/src/sdk/trace/span.dart
michaelyeager-wf: >I think we should add List.unmodifiable() to the attributes getter as well?

I would, but that emits an `Attributes`, not a List.   I think this is a good point to revisit during upcoming work on Link and Attribute classes.  I've filed tickets for both of those.

lib/src/sdk/trace/span_processors/span_processor.dart
blakeroberts-wk: Yeah, the spec docs have moved around a bit recently. Maybe the live pages have settled a bit at least 🤞 : https://opentelemetry.io/docs/specs/otel/trace/sdk/#span-processor

blakeroberts-wk (approved):

PR #127, opened 2023-10-04
Add deprecation messages for future changes to SpanProcessor and SpanExporter

michaelyeager-wf:

This PR adds deprecation messages for planned changes to SpanProcessor and SpanExporter.

## How Has This Been Tested?

* Confirmed unit tests pass successfully following this change.

### DIFF
--- a/lib/src/api/exporters/span_exporter.dart
+++ b/lib/src/api/exporters/span_exporter.dart
@@ -3,6 +3,7 @@
 
 import '../../../api.dart' as api;
 
+@Deprecated('This class will be moved to the SDK package in v0.17.0.')
 abstract class SpanExporter {
   void export(List<api.Span> spans);
 
--- a/lib/src/api/exporters/span_exporter.dart
+++ b/lib/src/api/exporters/span_exporter.dart
@@ -3,7 +3,8 @@
 
 import '../../../api.dart' as api;
 
-@Deprecated('This class will be moved to the SDK package in v0.17.0.')
+@Deprecated(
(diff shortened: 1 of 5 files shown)


semveraudit-wf:
### Public API Changes

 No changes to the public API found for commit 0447e3a85dca899eb98e65db0204ef6ea0b0844c

Showing results for 0447e3a85dca899eb98e65db0204ef6ea0b0844c
[Browse public API](https://semver-audit.wk-dev.wdesk.org/Workiva/opentelemetry-dart/0447e3a85dca899eb98e65db0204ef6ea0b0844c).

lib/src/api/exporters/span_exporter.dart
```diff
@@ -3,6 +3,7 @@
 
 import '../../../api.dart' as api;
 
+@Deprecated('This class will be moved to the SDK package in v0.17.0.')
 abstract class SpanExporter {
```
evanweible-wf, line 7: If possible, I always recommend moving these things (and leaving a copy or a re-export in the original location) _ahead_ of the major release so that consumers have a path to remediate the deprecation that isn't coupled to the major upgrade. It makes the upgrade process much smoother when consumer scan remediate deprecations ahead of time rather than all at once.

michaelyeager-wf: After some further thought, I think it'd be easier to move/deprecate these classes in https://github.com/Workiva/opentelemetry-dart/pull/125 for v.0.17 and remove them in another release after that.  I'll do that instead.

PR #129, opened 2023-10-24
O11Y-3111: Migrate OpenTelemetry Dart to null-safety

michaelyeager-wf:

This PR ensures null-safety for this library.

## How Has This Been Tested?

* Ensure unit tests pass successfully.
* Used a short test application to ensure these changes properly emit traces.


aviary-wf:
## Security Insights

## Action Items
---

.gitignore
```diff
@@ -1,3 +1,4 @@
 .dart_tool
 .packages
 pubspec.lock
+*.mocks.dart
```
michaelyeager-wf, line 4: This excludes generated test mocks.  Normally I'd commit these to the repo, but Dart's build runner gets cranky if "old" versions of the mocks exist.

See: https://ci.webfilings.com/build/5107873
For those unable to access the above:
```
dart run build_runner build  # Generate mocks for unit tests.  For config, see build.yaml.
[INFO] Generating build script...
[INFO] Generating build script completed, took 511ms
[INFO] Precompiling build script......
[INFO] Precompiling build script... completed, took 9.5s
[INFO] Initializing inputs
[INFO] Building new asset graph...
[INFO] Building new asset graph completed, took 1.9s
[INFO] Checking for unexpected pre-existing outputs....
[INFO] Found 1 declared outputs which already exist on disk. This is likely because the`.dart_tool/build` folder was deleted, or you are submitting generated files to your source repository.
Delete these files?
1 - Delete
2 - Cancel build
3 - List conflicts

You have hit a bug in build_runner
Please file an issue with reproduction steps at https://github.com/dart-lang/build/issues
Null check operator used on a null value
package:build_runner_core/src/environment/io_environment.dart 67:41           IOEnvironment.prompt
package:build_runner_core/src/environment/overridable_environment.dart 63:16  OverrideableEnvironment.prompt
package:build_runner_core/src/generate/build_definition.dart 546:41           _Loader._initialBuildCleanup
package:build_runner_core/src/generate/build_definition.dart 283:17           _Loader.prepareWorkspace.<fn>
package:build_runner_core/src/logging/logging.dart 25:30                      logTimedAsync
package:build_runner_core/src/generate/build_definition.dart 280:13           _Loader.prepareWorkspace
package:build_runner_core/src/generate/build_impl.dart 114:27                 BuildImpl.create
package:build_runner_core/src/generate/build_runner.dart 34:26                BuildRunner.create
package:build_runner/src/generate/build.dart 109:17                           build
package:build_runner/src/entrypoint/build.dart 35:18                          BuildCommand._run
package:args/command_runner.dart 212:13                                       CommandRunner.runCommand
package:build_runner/src/entrypoint/run.dart 26:18                            run
.dart_tool/build/entrypoint/build.dart 164:16                                 main
make: *** [Makefile:5: init] Error 1
```

I felt generating the mocks fresh each time was slightly better than a scripted solution to "cheese" terminal input during the build.

semveraudit-wf:
### Public API Changes

```diff
@@ opentelemetry/sdk.dart <-- package:opentelemetry/src/sdk/trace/exporters/collector_exporter.dart#L18 @@
class CollectorExporter implements SpanExporter
-     Uri uri
+     Uri get uri
//    Removed setter for `uri`.
//    Removing a field is a major change.

-     Map<String, String> headers
+     Map<String, String> get headers
//    Removed setter for `headers`.
//    Removing a field is a major change.

-     Client client
+     Client get client
//    Removed setter for `client`.
//    Removing a field is a major change.

```

```diff
@@ opentelemetry/sdk.dart <-- package:opentelemetry/src/sdk/trace/tracer_provider.dart#L14 @@
class TracerProviderBase implements TracerProvider
-     TracerProviderBase TracerProviderBase({List<SpanProcessor> processors, Resource resource, Sampler sampler, IdGenerator idGenerator, SpanLimits spanLimits})
+     TracerProviderBase TracerProviderBase({List<SpanProcessor> processors = const [], dynamic resource, Sampler sampler = const sdk.ParentBasedSampler(sdk.AlwaysOnSampler()), IdGenerator idGenerator = const sdk.IdGenerator(), SpanLimits spanLimits = const sdk.SpanLimits()})
//    `type` of `resource` has changed.
//    Changing a parameter signature is a major change.

```

```diff
@@ opentelemetry/sdk.dart <-- package:opentelemetry/src/sdk/trace/sampling/parent_based_sampler.dart#L10 @@
class ParentBasedSampler implements Sampler
-     ParentBasedSampler ParentBasedSampler(Sampler _root, {dynamic remoteParentSampled, dynamic remoteParentNotSampled, dynamic localParentSampled, dynamic localParentNotSampled})
+     ParentBasedSampler ParentBasedSampler(Sampler _root, {Sampler remoteParentSampled = const sdk.AlwaysOnSampler(), Sampler remoteParentNotSampled = const sdk.AlwaysOffSampler(), Sampler localParentSampled = const sdk.AlwaysOnSampler(), Sampler localParentNotSampled = const sdk.AlwaysOffSampler()})
//    `type` of `localParentNotSampled` has changed.
//    Changing a parameter signature is a major change.

+     Sampler get remoteParentNotSampled
//    Adding a field is a minor change.

+     Sampler get localParentNotSampled
//    Adding a field is a minor change.

+     Sampler get localParentSampled
//    Adding a field is a minor change.

+     Sampler get remoteParentSampled
//    Adding a field is a minor change.

```

Showing results for 65aae2cffbfe63fe97c0a56e495621fb0d34293f
[Browse public API](https://semver-audit.wk-dev.wdesk.org/Workiva/opentelemetry-dart/65aae2cffbfe63fe97c0a56e495621fb0d34293f).

lib/src/sdk/common/limits.dart
```diff
@@ -0,0 +1,80 @@
+// Copyright 2021-2022 Workiva.
+// Licensed under the Apache License, Version 2.0. Please see https://github.com/Workiva/opentelemetry-dart/blob/master/LICENSE for more information
+import 'package:meta/meta.dart';
+
+import '../../../api.dart' as api;
+import '../../../sdk.dart' as sdk;
+
+@protected
+List<api.SpanLink> applyLinkLimits(
```
michaelyeager-wf, line 9: I recognize these may not need to be or ultimately end up staying here.

I've split these out as "utility" methods as we do more work on Links, Attributes, and limits in general.  Pulling these methods out of Span:
* Allows a Tracer to impose limits when creating a Span.  This simplifies the construction of a Span and allows us to get closer to `const`-ing SDK Span's constructor.
* Allows a Span to impose limits to attributes added to it.

As I said, we have work coming up to improve how we handle Links, Attributes, and limits.  This seemed at least _a_ middle-ground as we explore those changes.  I'd welcome some discussion here, as this part of the library is still pretty rough.

lib/src/api/context/context.dart
```diff
@@ -76,11 +76,11 @@ class Context {
 
   /// Get the [api.Span] attached to this [Context], or null if no such
   /// [api.Span] exists.
-  api.Span get span => getValue(spanKey);
+  api.Span? get span => getValue(spanKey);
```
blakeroberts-wk, line 79: should this return a non-recording span?

lib/src/api/context/context.dart
```diff
@@ -76,11 +76,11 @@ class Context {
...
 
   /// Get the [api.SpanContext] from this [Context], or null if no such
   /// [api.SpanContext] exists.
-  api.SpanContext get spanContext => getValue(spanKey)?.spanContext;
+  api.SpanContext? get spanContext => getValue(spanKey)?.spanContext;
```
blakeroberts-wk, line 83: should this return an invalid span context?

lib/src/api/propagation/extractors/text_map_getter.dart
```diff
@@ -9,5 +9,5 @@ abstract class TextMapGetter<C> {
   Iterable<String> keys(C carrier);
 
   /// Returns the first value of the given propagation [key] or returns null.
-  String get(C carrier, String key);
+  String? get(C carrier, String key);
```
blakeroberts-wk, line 12: should this return empty string instead of null?

lib/src/api/trace/span_id.dart
```diff
@@ -9,7 +9,7 @@ class SpanId {
   static const sizeBits = 16;
   static const sizeBytes = 8;
 
-  List<int> _id;
+  late List<int> _id;
```
blakeroberts-wk, line 12: can this be marked final?

lib/src/api/trace/trace_id.dart
```diff
@@ -9,7 +9,7 @@ class TraceId {
   static const sizeBits = 32;
   static const sizeBytes = 16;
 
-  List<int> _id;
+  late List<int> _id;
```
blakeroberts-wk, line 12: can this be marked final?

lib/src/api/trace/trace_state.dart
```diff
@@ -70,7 +70,7 @@ class TraceState {
     return (match != null) && (match.group(0) == value);
   }
 
-  String get(String key) => _state[key];
+  String? get(String key) => _state[key];
```
blakeroberts-wk, line 73: should this return empty string if not found?
```suggestion
  String get(String key) => _state[key] ?? '';
```

lib/src/sdk/resource/resource.dart
```diff
@@ -5,16 +5,10 @@ import '../../../api.dart' as api;
...
 
-  Resource(List<api.Attribute> attributes) {
-    for (final attribute in attributes) {
-      if (attribute.value is! String) {
-        throw ArgumentError('Attributes value must be String.');
-      }
-    }
-    _attributes.addAll(attributes);
-  }
+  Resource(List<api.Attribute> attributes)
```
blakeroberts-wk, line 10: can this be marked a constant constructor?

lib/src/sdk/trace/exporters/collector_exporter.dart
```diff
@@ -139,13 +138,13 @@ class CollectorExporter implements sdk.SpanExporter {
     return pb_trace.Span(
         traceId: span.spanContext.traceId.get(),
         spanId: span.spanContext.spanId.get(),
-        parentSpanId: span.parentSpanId?.get(),
+        parentSpanId: span.parentSpanId.get(),
         name: span.name,
         startTimeUnixNano: span.startTime,
         endTimeUnixNano: span.endTime,
         attributes: span.attributes.keys.map((key) => pb_common.KeyValue(
```
blakeroberts-wk, line 145:
```suggestion
        attributes: span.attributes.entries.map((key) => pb_common.KeyValue(
```

lib/src/sdk/trace/sampling/parent_based_sampler.dart
```diff
@@ -1,27 +1,31 @@
...
-      localParentSampled,
-      localParentNotSampled})
-      : _remoteParentSampled = remoteParentSampled ?? sdk.AlwaysOnSampler(),
-        _remoteParentNotSampled =
-            remoteParentNotSampled ?? sdk.AlwaysOffSampler(),
-        _localParentSampled = localParentSampled ?? sdk.AlwaysOnSampler(),
-        _localParentNotSampled =
-            localParentNotSampled ?? sdk.AlwaysOffSampler();
+
+  @protected
```
blakeroberts-wk, line 12: are these fields needed outside of this package or can they be private?

lib/src/sdk/trace/span.dart
```diff
@@ -22,10 +24,12 @@ class Span implements sdk.ReadWriteSpan {
   final sdk.InstrumentationScope _instrumentationScope;
   final Int64 _startTime;
   final Attributes _attributes = Attributes.empty();
+
   String _name;
-  Int64 _endTime;
   int _droppedSpanAttributes = 0;
 
+  Int64? _endTime;
```
blakeroberts-wk, line 31: can this be final?

lib/src/sdk/trace/span.dart
```diff
@@ -177,70 +150,16 @@ class Span implements sdk.ReadWriteSpan {
 
   @override
   void addEvent(String name, Int64 timestamp,
-      {List<api.Attribute> attributes}) {
+      {List<api.Attribute>? attributes}) {
```
blakeroberts-wk, line 153: can this get a default value?

lib/src/sdk/trace/span.dart
```diff
@@ -46,30 +50,16 @@ class Span implements sdk.ReadWriteSpan {
       this._timeProvider,
       this._resource,
       this._instrumentationScope,
-      api.SpanKind kind,
-      List<api.Attribute> attributes,
```
blakeroberts-wk, line 50: I'm not sure we want to omit the attributes argument from the constructor

lib/src/sdk/trace/span.dart
```diff
@@ -79,15 +69,15 @@ class Span implements sdk.ReadWriteSpan {
 
   @override
   void end() {
-    _endTime ??= _timeProvider.now;
+    _endTime = _timeProvider.now;
```
blakeroberts-wk, line 72: after the first call to end, any additional call should no-op

lib/src/sdk/trace/span_limits.dart
```diff
@@ -90,12 +90,12 @@ class SpanLimits {
   ///https://docs.newrelic.com/docs/data-apis/manage-data/view-system-limits/
   ///https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/SpanLimitsBuilder.java
```
blakeroberts-wk, line 91: remove link

lib/src/sdk/trace/tracer.dart
```diff
@@ -26,16 +27,19 @@ class Tracer implements api.Tracer {
...
   api.Span startSpan(String name,
-      {api.Context context,
-      api.SpanKind kind,
-      List<api.Attribute> attributes,
-      List<api.SpanLink> links,
-      Int64 startTime}) {
+      {api.Context? context,
+      api.SpanKind kind = api.SpanKind.internal,
+      List<api.Attribute>? attributes,
+      List<api.SpanLink>? links,
```
blakeroberts-wk, line 37:
```suggestion
      List<api.Attribute> attributes = [],
      List<api.SpanLink> links = [],
```

lib/src/sdk/trace/tracer.dart
```diff
@@ -26,16 +27,19 @@ class Tracer implements api.Tracer {
...
+      this._limits);
 
   @override
   api.Span startSpan(String name,
-      {api.Context context,
-      api.SpanKind kind,
-      List<api.Attribute> attributes,
-      List<api.SpanLink> links,
-      Int64 startTime}) {
+      {api.Context? context,
```
blakeroberts-wk, line 34: If a context isn't constant, can it be made constant? or at least can there be an empty/invalid constant constructor to use as a default value here and elsewhere?

lib/src/sdk/trace/span_processors/batch_processor.dart
```diff
@@ -89,13 +89,10 @@ class BatchSpanProcessor implements sdk.SpanProcessor {
   }
 
   void _clearTimer() {
-    if (_timer == null) {
```
blakeroberts-wk, line 92: we should clean this code up:
1. remove `_clearTimer()` and `_startTimer()` and therefore any invocation
2. mark `_timer` as final and assign at construction
3. cancel `_timer` on shutdown only
4. remove `_addToBuffer()` moving its logic to `onEnd()` where it's called

michaelyeager-wf: I agree.  I explored exactly what you're suggesting during work on this set of changes, but ultimately decided against it.

Some context:
* A `Timer` may be `start`-ed or `cancel`-ed, but not paused.  
* Once `cancel`-ed, a `Timer` may not be restarted and should be discarded.

Thanks to these limitations, the implementation here was written in order to prevent the time spent sending spans from counting against the next batch interval.  I remember this being a very important point when this code first went in.

I couldn't reliably anticipate how changes here would perform under heavy use, though I could easily imagine a situation where emitting many spans might back up and overwhelm the processor.   I decided it best to wait for more opinions at least.

lib/src/sdk/trace/tracer.dart
michaelyeager-wf: >If a context isn't constant, can it be made constant?

I explored this.  In short, `Context` is wrapper around `Zone`.  Since `Zone` is not constant, `Context` cannot be either, since arguments defaulted in a constant constructor must themselves be constant.

>can there be an empty/invalid constant constructor

Only if such a constructor returns a `Context` which does not wrap a `Zone` for the reasons described above.  Since `Context` uses `Zone` for span storage, I don't know how such a thing could be used without causing problems.

lib/src/sdk/trace/tracer.dart
michaelyeager-wf: I explored this too.  Default optional values must be constant.  Constant `List`s may not be modified after creation, meaning any span created this way cannot have links or attributes added to it, violating the specification.

lib/src/sdk/trace/span.dart
michaelyeager-wf: Any final field must be initialized either at definition or construction.  We can't use a final field here because the timestamp is assigned at some time in the future.  I settled upon nullable because I didn't feel that some arbitrary "default time" was any better for determining whether the span had an end time than a null value was.

lib/src/sdk/resource/resource.dart
michaelyeager-wf: I explored this.  `Resource` cannot be constant because it wraps an `Attributes` for storage and `Attributes` wraps a `Map`.  `const`-ing `Resource` would require `Attributes` to be constant and thereby use a constant `Map`.  A constant `Map` cannot be modified after creation, meaning spans couldn't add attributes after creation either.  

Attributes themselves are a mess and we've created some tickets to resolve that.  I think untangling this situation might be better done there.

lib/src/api/trace/trace_state.dart
michaelyeager-wf: I'm not sure about this.  The current implementation is able to differentiate between "the state does not contain this key" and "the value for this key is an empty string".  I'm concerned that using a blank string here could allow blank values to leak into the trace state.

lib/src/api/context/context.dart
michaelyeager-wf: I'll take this question in tandem with the question about span context below, since they're mostly the same answer.

I'm not sure about this. The current implementation is able to differentiate between "the context does not contain a span" and "the context contains a span from a remote source".  We should be careful about disrupting that distinction.  I'm mostly concerned that the current remainder of the library might emit disconnected spans if "empty Context" defaults to spans with non-recording span parents.  I can make a branch to test this, if you'd like.

lib/src/api/trace/span_id.dart
michaelyeager-wf: I'll take this question in tandem with the question about `TraceId` below, since they're similar answers.

Final fields don't have setters, which we need for the `fromString` constructor below.  I think `late` is the way to go here because:
* It doesn't allow a null `_id_` value.
* Every existing constructor assigns a value to `_id`, ensuring that using a `late _id` won't throw.

lib/src/sdk/common/limits.dart
```diff
@@ -0,0 +1,80 @@
...
+    attr = api.Attribute.fromStringList(attr.key, listString);
+  }
+  return attr;
+}
+
+//Truncate just strings which length is longer than configuration.
+//Reference: https://github.com/open-telemetry/opentelemetry-java/blob/14ffacd1cdd22f5aa556eeda4a569c7f144eadf2/sdk/common/src/main/java/io/opentelemetry/sdk/internal/AttributeUtil.java#L80
+@protected
+String applyAttributeLengthLimit(String value, int lengthLimit) {
+  return value.length > lengthLimit ? value.substring(0, lengthLimit) : value;
```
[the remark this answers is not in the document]
michaelyeager-wf: Unfortunately, String doesn't have a `characters` getter in Dart.

lib/src/sdk/trace/exporters/collector_exporter.dart
```diff
@@ -68,7 +66,8 @@ class CollectorExporter implements sdk.SpanExporter {
       for (final attr in il.key.attributes.keys) {
```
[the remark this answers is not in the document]
michaelyeager-wf: We can't do this, because the `attributes` getter returns an `Attributes`, not a `Map`, and `Attributes` doesn't expose the `entries` method from its wrapped `Map`.  I think this is a good point to put a pin in for upcoming work on `Attributes`.

lib/src/sdk/trace/exporters/collector_exporter.dart
michaelyeager-wf: We can't do this, because the `attributes` getter returns an `Attributes`, not a `Map`, and `Attributes` doesn't expose the `entries` method from its wrapped `Map`.  I think this is a good point to put a pin in for upcoming work on `Attributes`.

.gitignore
michaelyeager-wf: I think I've found the solution here:

```
--delete-conflicting-outputs
```

This flag:

>By default, the user will be prompted to delete any files which already exist but were not known to be generated by this specific build script.  Enabling this option skips the  prompt and deletes the files. This should typically be used in continues integration servers and tests, but not otherwise.

This is our exact use case, so I've added this flag to our build command.

lib/src/sdk/trace/span.dart
michaelyeager-wf: I can move this parameter back in, if you want.  However, I reasoned that:
* These spans are (should be) only created by `Tracer` within the SDK.
* `Tracer` can just as easily use `Span.setAttributes` when creating a span, since this is what Span's constructor was already doing.

lib/src/sdk/trace/span.dart
michaelyeager-wf: Good call.  There's no reason an event's attributes should change after they're added.

lib/src/sdk/trace/span_processors/batch_processor.dart
blakeroberts-wk: p99 on sending spans tends to be below 100ms, and the default client timeout is 10 seconds. generally, the interval is going to be 60 seconds. Even if the time spent sending spans could be counted against the timer, I'm not sure it'd be a relevant concern. Moreover, the timer is async so we could fire off an async send event without awaiting to avoid this potential issue.

I suppose these changes are out of scope for this work though. Could we make a ticket to follow up on this at a later date?

lib/src/api/trace/span_link.dart
```diff
@@ -7,7 +7,7 @@ class SpanLink {
   final api.SpanContext _context;
   final List<api.Attribute> _attributes;
 
-  SpanLink(this._context, {List<api.Attribute> attributes})
+  SpanLink(this._context, {List<api.Attribute>? attributes})
       : _attributes = attributes ?? [];
```
blakeroberts-wk, line 11: In dart, does having a question mark on the optional named parameter mean that doing `new SpanLink(ctx, {attributes: null})` is valid? Can we avoid this by doing something like the following?

```suggestion
  SpanLink(this._context, {List<api.Attribute> attributes = const []})
      : _attributes = new List<api.Attribute>()..addAll(attributes);
```

lib/src/sdk/trace/span.dart
blakeroberts-wk: can the field be `late final`?

lib/src/sdk/trace/span.dart
blakeroberts-wk: That's fair

lib/src/sdk/metrics/state/meter_shared_state.dart
```diff
@@ -12,7 +12,7 @@ class MeterSharedState {
   final MeterProviderSharedState _meterProviderSharedState;
   // ignore: unused_field
   final InstrumentationScope _instrumentationScope;
-  sdk.Meter meter;
+  late sdk.Meter meter;
```
robbecker-wf, line 15: You could avoid this late by using an initialize in the constructor like
```dart
  MeterSharedState(this._meterProviderSharedState, this._instrumentationScope) : meter = sdk.Meter(this);
```

lib/src/sdk/trace/exporters/collector_exporter.dart
```diff
@@ -13,14 +14,15 @@ import '../../proto/opentelemetry/proto/resource/v1/resource.pb.dart'
...
+  final Map<String, String> headers;
   var _isShutdown = false;
 
-  CollectorExporter(this.uri, {http.Client httpClient, this.headers}) {
-    client = httpClient ?? http.Client();
-  }
+  CollectorExporter(this.uri,
+      {http.Client? httpClient, Map<String, String>? headers})
+      : client = httpClient ?? http.Client(),
+        headers = headers ?? {};
```
robbecker-wf, line 25: I think you could shorten this up to:
```suggestion
  CollectorExporter(this.uri,
      {http.Client? httpClient, this.headers = const{}})
      : client = httpClient ?? http.Client();
```

lib/src/sdk/trace/exporters/collector_exporter.dart
```diff
@@ -68,7 +66,8 @@ class CollectorExporter implements sdk.SpanExporter {
       for (final attr in il.key.attributes.keys) {
         attrs.add(pb_common.KeyValue(
             key: attr,
-            value: _attributeValueToProtobuf(il.key.attributes.get(attr))));
+            value:
+                _attributeValueToProtobuf(il.key.attributes.get(attr) ?? '')));
```
robbecker-wf, line 70: Is this a logic change? previously it's possible to get a null back from the attributes.get call .. and now you're guaranteed to get at least an empty string. I know that's the intention to make it non-nullable, but is that OK? Will it change any behavior?

lib/src/sdk/trace/exporters/console_exporter.dart
```diff
@@ -15,7 +15,7 @@ class ConsoleExporter implements sdk.SpanExporter {
         'name': span.name,
         'id': '${span.spanContext.spanId}',
         'timestamp': span.startTime,
-        'duration': span.endTime - span.startTime,
+        'duration': span.endTime! - span.startTime,
```
robbecker-wf, line 18: Is the `!` what you want here? If the span is still running, then endTime will be null and this will throw. Maybe it should handle that null case without an exception.

lib/src/sdk/trace/span.dart
robbecker-wf: I agree that a null here signifies something ... that it hasn't ended yet.

lib/src/sdk/trace/span_limits.dart
```diff
@@ -86,23 +86,17 @@ class SpanLimits {
...
-        _maxNumAttributesPerEvent =
-            maxNumAttributesPerEvent ?? _DEFAULT_MAXNUM_ATTRIBUTE_PER_EVENT,
-        _maxNumAttributesPerLink =
-            maxNumAttributesPerLink ?? _DEFAULT_MAXNUM_ATTRIBUTES_PER_LINK,
-        _maxNumAttributeLength =
-            maxNumAttributeLength ?? _DEFAULT_MAXNUM_ATTRIBUTES_LENGTH;
+      {int maxNumAttributes = _DEFAULT_MAXNUM_ATTRIBUTES,
+      int maxNumEvents = _DEFAULT_MAXNUM_EVENTS,
+      int maxNumLink = _DEFAULT_MAXNUM_LINKS,
+      int maxNumAttributesPerEvent = _DEFAULT_MAXNUM_ATTRIBUTE_PER_EVENT,
```
robbecker-wf, line 93: Should ATTRIBUTES be plural here too?

lib/src/sdk/trace/span_processors/batch_processor.dart
robbecker-wf: FYI there's a restartable timer https://pub.dev/documentation/async/latest/async/RestartableTimer-class.html

lib/src/api/trace/span_link.dart
robbecker-wf: Oh, I see that you may want to be able to add attributes later which using a const list would prevent. Feel free to disregard my suggestion.

pubspec.yaml
```diff
@@ -3,7 +3,7 @@ version: 0.16.2
 description: A framework for collecting traces from applications.
 homepage: https://github.com/Workiva/opentelemetry-dart
 environment:
-  sdk: ">=2.11.0 <3.0.0"
+  sdk: ">=2.15.0 <4.0.0"
 
 dependencies:
   collection: ^1.0.0
```
robbecker-wf, line 9: Since we're declaring null safety, then all dependencies in your pubspec should have their minimums raised to at least the first nullsafe version. For example, collection ^1.15.0 (see the version listing for a package https://pub.dev/packages/collection/versions)

The same applies to fixnum, http, etc.

pubspec.yaml
```diff
@@ -3,7 +3,7 @@ version: 0.16.2
 description: A framework for collecting traces from applications.
 homepage: https://github.com/Workiva/opentelemetry-dart
 environment:
-  sdk: ">=2.11.0 <3.0.0"
+  sdk: ">=2.15.0 <4.0.0"
```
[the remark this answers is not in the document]
michaelyeager-wf: >Is there a reason you chose 2.15 as the minimum here?
2.15 was a requirement for the build_[runner/test/web_compiler] dependencies added below.  I suppose we could go higher, but I didn't want to get too prescriptive if I could help it.

>Do we properly test this library in Dart 3?
No, so I agree that we should turn this back down.

lib/src/sdk/trace/span.dart
michaelyeager-wf: >can the field be late final?

I don't think we should do that.  With a `late` variable if we attempt to access `_endTime` before we assign an end time we'll get a runtime error.

blakeroberts-wk: I agree with having null signify the span hasn't ended. I was hoping that it'd be possible to enforce the field is assigned to only once. But it seems neither final, nor late final is sufficient for this use-case

lib/src/sdk/common/limits.dart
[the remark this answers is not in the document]
blakeroberts-wk: Ah. It's an extension: https://pub.dev/documentation/characters/latest/characters/StringCharacters.html

lib/src/api/context/context.dart
blakeroberts-wk: I believe a context that does not contain a span should be invalid whereas a context that contains a span from a remote source should be valid and either of which is non-recording. Or maybe I am misremembering spec. Either way, returning a non-recording span here is nice so that code can do something like `otel.SpanFromContext(ctx).SetAttributes(...)` without worrying about nulls. Though in Dart/TS/etc, it isn't a big deal since callers only need to add a `?` between the two calls

lib/src/api/trace/trace_state.dart
blakeroberts-wk: https://opentelemetry.io/docs/specs/otel/trace/tracestate-handling

The grammar suggests that empty values are valid values. And based on this document, I do not think it is worth marking this APIs return value as `String?` because afaict `String` should be sufficient and in the case `String?` must change to `String` that is a breaking change. But going from `String` to `String?` is not a breaking change

lib/src/sdk/trace/sampling/parent_based_sampler.dart
```diff
@@ -1,27 +1,35 @@
...
 
+import 'package:meta/meta.dart';
+
 import '../../../../api.dart' as api;
 import '../../../../sdk.dart' as sdk;
 
 class ParentBasedSampler implements sdk.Sampler {
   final sdk.Sampler _root;
+
+  @protected
```
blakeroberts-wk, line 12: Does `@protected` modify the scope past what appending `_` does?

robbecker-wf: They are different



private https://dart.dev/language/libraries
> identifiers that start with an underscore (_) are visible only inside the library. Every Dart file (plus its parts) is a [library](https://dart.dev/tools/pub/glossary#library), even if it doesn’t use a [library](https://dart.dev/language/libraries#library-directive) directive.

vs

protected https://pub.dev/documentation/meta/latest/meta/protected-constant.html
> Used to annotate an instance member in a class or mixin which is meant to be visible only within the declaring library, **and to other instance members of the class or mixin, and their subtypes.**

lib/src/sdk/trace/tracer.dart
blakeroberts-wk: 👍 sounds good. I have a sneaky suspicion this library will one day remove its reliance on zones. But that's neither here, nor there.

lib/src/sdk/trace/tracer.dart
blakeroberts-wk: Is it possible to use the constant and then destructure and create the arg in assignment to the member? Something like `_attributes = [...attributes]`?

Main concern is avoiding the ability to explicitly assign null to the named optional parameter. Which might not be possible in Dart, not sure. But the underlying implication here can have some very awkward bugs crop up in other languages like Go.

lib/src/api/trace/trace_state.dart
michaelyeager-wf: >The grammar suggests that empty values are valid values.

Yes, but that's not what I'm concerned about.  Do we want the library to start returning valid data for keys that don't exist?

lib/src/sdk/trace/tracer.dart
michaelyeager-wf: >Main concern is avoiding the ability to explicitly assign null to the named optional parameter.

This is great info to know and now your question make a _lot_ more sense.  

1. Lines 40-41 should handle null values passed in.
2. Otherwise, we'd have to do like `List<api.Attribute> attributes = const []` instead, since default values _must_ be constant.

We have to be careful about how those values are used, since constant lists can't be modified.  Taking a glance of how Tracer uses these values, we're going to get off easy.  The attributes and links are added to the span through `setAttributes`/`applyLInkLimits`, which should apply limits and reassign each item into the span, _not_ set a reference to that constant list.  

Unit tests show default values working, I'll get this committed.

lib/src/sdk/trace/sampling/parent_based_sampler.dart
michaelyeager-wf: These are leftovers not cleaned up after some of the earlier requested changes, which saw these changed from protected class fields to private class fields.  These can be removed.

lib/src/api/context/context.dart
michaelyeager-wf: I don't remember that the spec is prescriptive either way.  Like I said, I can test this.  We need to be careful that root spans don't end up with a non-recording trace parent that "doesn't exist".

lib/src/api/trace/trace_state.dart
blakeroberts-wk: I would not recommend that instrumentation leveraging `tracestate` define null and empty string as two distinct states. I would consider this bad practice.

But this is just my opinion. Perhaps there's a valid reason for treating the two as distinct states. There isn't, but someone could make a case for it. At which point we could add `?` to the API and that change is backwards compatible.

Here's some pseudo code that pretends to use trace state that can distinguish between null and empty value:

```dart
const value = ctx.get('my-key')

if (value == null) {
// initialize value to something non-empty
} else if (value == "") {
// this is a special case that implies something special?
// or more likely we treat it the same as null and initialize the value to something non-empty
} else {
// value is non-null and non-empty
}
```

lib/src/api/propagation/extractors/text_map_getter.dart
michaelyeager-wf: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/api-propagators.md#get

robbecker-wf:
@aviary-wf

lib/src/sdk/metrics/state/meter_shared_state.dart
robbecker-wf: either way this late is safe since the constructor always initializes it.

robbecker-wf (approved):

lib/src/sdk/trace/exporters/collector_exporter.dart
michaelyeager-wf: Not really, attributes is a map under the hood, so it's more to get the analyzer to shut up.  This iterates over the existing keys in the first place, so any key encountered here must exist.

lib/src/api/propagation/extractors/text_map_getter.dart
blakeroberts-wk: makes sense. imo there's def a stronger case to have this return null than tracestate. And if we'd like to keep API consistent between the two I suppose this seems reasonable

opentelemetry-dart.iml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
blakeroberts-wk, line 1: remove

lib/src/sdk/metrics/state/meter_shared_state.dart
blakeroberts-wk: I think this code might need to be turned around a bit 🤔 I'm guessing the code that manages `MeterSharedState` should probably manage the `Meter` directly w/o this middleman therefore avoiding the need to pass `this` to a non-super constructor within a constructor

michaelyeager-wf: >the code that manages MeterSharedState should probably manage the Meter directly

That's `MeterProvider`, so absolutely it should.

Do you want me to pull this apart here, or ticket it for later?

blakeroberts-wk: Ticket for later

lib/src/sdk/trace/exporters/collector_exporter.dart
blakeroberts-wk: Does that mean we should use `!`?
```suggestion
                _attributeValueToProtobuf(il.key.attributes.get(attr)!)));
```

blakeroberts-wk (approved):

PR #136, opened 2023-11-16
Implemented Span.addEvent and added tests for it

### LINKED ISSUE #116: Support for Events?
**Is your feature request related to a problem? Please describe.**
The library does not support Events (throws [UnimplementedError](https://github.com/Workiva/opentelemetry-dart/blob/815985946cd7c2d1be3423818d3d988a12183173/lib/src/sdk/trace/span.dart#L169))

**Describe the solution you'd like**
I'd like to use events to track pageviews and non-fatal exceptions/traces.
For a Flutter application, tracking these 2 I think makes sense in order to investigate potential issues. It is not enough to see an HTTP call to the server, without knowing the page flow of the user or if there were any exceptions on the client side.
Thats what Open Telemetry is for, at least in my mind: to give you a clear view of what is going on inside the components of your production.

**Describe alternatives you've considered**
I have considered "recordException" or attributes, but this will only record the latest value of each key. This is not helpful as there may be multiple "pages" or "traces" in a single span.
I have also considered new spans for each trace, but that is suboptimal, given that most SaaS collectors charge per span.
Unfortunatelly this is what I have to use for now.

**Additional context**
Awsome effort, looking forward to events and nullability support, thank you so much!
PS: I am using the opentelemetry_sls: ^0.15.1 package, because your package does not support Dart 3.

baskillen:

Adding proper support for Span.addEvent which was previously throwing an UnimplementedError

Fixes #116 

- Implements the Span.addEvent function.
- Added an api model SpanEvent and added a list of events in Span.
- Includes the events in the protobuf request in the collector_exporter.
- Adds information to the protobuf even on how many events that were dropped (due to SpanLimits)
- Did some deduplication in Span_test.dart and added test for adding events.
- Improved the comparison in collector_exporter_test to actually understand what differs.

## How Has This Been Tested?

- Ran all the unit tests

## Checklist:


aviary2-wf:
## Security Insights

## Action Items
---

StephenWithPH:
@baskillen ... this is great! I'll happily await this merge. 👏

lib/src/api/trace/span.dart
```diff
@@ -60,7 +60,8 @@ abstract class Span {
   void setAttributes(List<api.Attribute> attributes);
 
   /// Record metadata about an event occurring during this span.
-  void addEvent(String name, Int64 timestamp, {List<api.Attribute> attributes});
+  void addEvent(String name,
+      {Int64? timestamp, List<api.Attribute> attributes});
```
blakeroberts-wk, line 64: Because this is a breaking change, this will need to be released in version v0.18.0. Right now, any merge to master constitutes a release of this project. We might revise this strategy at a later date, but for now this is where we are.

With this being said, which route for release of this feature would we prefer:
1. This PR will constitute release v0.18.0 (might take a moment to get this published and may not have an ETA).
2. We avoid breaking changes by keeping the API as it was and have this PR release as a patch increment. We can add a deprecation warning about breaking the method's definition in a future release (v0.18.0). This'll make releasing easier therefore faster. But the API will be awkward to use and its use brittle as it will break in the future.

I have no strong opinions either way.

blakeroberts-wk (commented):
Thanks for the PR! The changes look good. I have only one awkward question about how we'd like to release this feature given that it is a breaking change in its current state (which is understandable and acceptable).

lib/src/api/trace/span.dart
StephenWithPH: For what it's worth, I believe the only concrete implementation of this currently throws an Error. So I suppose this change could cause compile errors, but only in the case where that code path would have caused a runtime error had it been reached.

This seems acceptable (to me at least) when combined with the the fact that this change brings the method into compliance with [the spec](https://opentelemetry.io/docs/specs/otel/trace/api/#add-events).

https://github.com/Workiva/opentelemetry-dart/blob/322f988a3bc43a2ffe8dff1973e8983c61dedcf1/lib/src/sdk/trace/span.dart#L155-L160

blakeroberts-wk: Doh! Good call. Thank you for this information. With this, I am supremely in-favor of option 1.

blakeroberts-wk (approved):

baskillen:
> Thanks for the PR! The changes look good. I have only one awkward question about how we'd like to release this feature given that it is a breaking change in its current state (which is understandable and acceptable).

Thanks for responding so quickly. To me option 1 sounds good. We will use our local copy until we have 0.18 in place.
Any idea why the test build step is not running?

lib/src/api/trace/span.dart
baskillen: Thanks for engaging @StephenWithPH! ❤️

StephenWithPH:
@blakeroberts-wk ...

It looks like the `test` GitHub Actions workflow needs approval to run: https://github.com/Workiva/opentelemetry-dart/actions/runs/6891390980

StephenWithPH:
Checking in... does this PR need anything else to be ready to merge? I see the `Merge Requirements Unmet` label remains.

blakeroberts-wk:
QA +1 passing unit tests are sufficient for these changes

micahbeeman-wf:
RM +1

Issue #101, opened 2023-02-20
Does the CollectorExporter support gRPC?


cnbleu:
Does the [CollectorExporter](https://github.com/Workiva/opentelemetry-dart#collectorxxporter) support gRPC like [[OTLP Exporters](https://github.com/open-telemetry/opentelemetry-java/blob/main/exporters/otlp/all)](https://github.com/open-telemetry/opentelemetry-java/tree/main/exporters/otlp/all) in Java language?

I hava configured the CollecctorExporter with the gRPC url, but got below errors:

![image](https://user-images.githubusercontent.com/3634065/220065858-c1c86ebf-96e7-4ca8-94bd-bf3478b6de74.png)

If it supports gRPC, any examples are expected.

cnbleu:
Hi @bryonmarks-wf , are you the person to ask?

bryonmarks-wf:
Hey @cnbleu, I'm not.  @blakeroberts-wk is this something you could help with?

blakeroberts-wk:
The library does not support gRPC at this time.

cnbleu:
> The library does not support gRPC at this time.

Hi @blakeroberts-wk. Is there any relevant plan support gRPC?

blakeroberts-wk:
> Is there any relevant plan support gRPC?

There are currently no plans to support gRPC; however, support is within spec and should be a long term goal of the project.

PR #161, opened 2024-03-27
Feature: Support gRPC protocol in CollectorExporter

huandu:

In this PR, add gRPC support in `CollectorExporter`.

Fixes # (issue)

## How Has This Been Tested?

I've used this client in my production environment to export traces to Aliyun SLS, which supports gPRC protocol only.

## Checklist:


aviary2-wf:
## Security Insights

## Action Items
---

huandu:
This PR should be a fix for #101.

lib/src/sdk/trace/exporters/collector_exporter.dart
```diff
@@ -48,10 +67,24 @@ class CollectorExporter implements sdk.SpanExporter {
     try {
       final body = pb_trace_service.ExportTraceServiceRequest(
           resourceSpans: _spansToProtobuf(spans));
-      final headers = {'Content-Type': 'application/x-protobuf'}
-        ..addAll(this.headers);
 
-      await client.post(uri, body: body.writeToBuffer(), headers: headers);
+      switch (protocol) {
```
blakeroberts-wk, line 71: What if we avoided the switch statement using an abstract client member assigned a concrete value at initialization of this class? Something like the following:

```dart
enum CollectorExporterProtocol { http, gRPC }

abstract class _CollectorExporterClient {
  Future<void> send(pb_trace_service.ExportTraceServiceRequest body,
      Map<String, String> headers);

  void close();
}

class _HTTPExporterClient implements _CollectorExporterClient {
  static final Logger _log = Logger('opentelemetry._HTTPExporterClient');

  final Uri _uri;
  final http.Client _client;

  const _HTTPExporterClient(this._uri, this._client);

  @override
  Future<void> send(pb_trace_service.ExportTraceServiceRequest body,
      Map<String, String> headers) async {
    try {
      await _client.post(_uri,
          body: body.writeToBuffer(),
          headers: {'Content-Type': 'application/x-protobuf'}..addAll(headers));
    } catch (e) {
      _log.warning('Failed to export spans.', e);
    }
  }

  @override
  void close() {
    _client.close();
  }
}

class _GRPCExporterClient implements _CollectorExporterClient {
  static final Logger _log = Logger('opentelemetry._GRPCExporterClient');

  final dynamic _client;

  const _GRPCExporterClient(this._client);

  @override
  Future<void> send(pb_trace_service.ExportTraceServiceRequest body,
      Map<String, String> headers) async {
    try {
      await _client.export(
        body,
        options: grpc.CallOptions(
          metadata: headers,
        ),
      );
    } catch (e) {
      _log.warning('Failed to export spans.', e);
    }
  }

  @override
  void close() {}
}

class CollectorExporter implements sdk.SpanExporter {
  @Deprecated('This member will be removed in v0.19.0.')
  final Uri uri;

  @Deprecated('This member will be removed in v0.19.0.')
  final http.Client client;
  final _CollectorExporterClient _client;

  @Deprecated('This member will be removed in v0.19.0.')
  final Map<String, String> headers;
  final Map<String, String> _headers;

  late bool _isShutdown;

  CollectorExporter(this.uri,
      {http.Client? httpClient,
      this.headers = const {},
      protocol = CollectorExporterProtocol.http})
      : client = httpClient ?? http.Client(),
        _headers = headers,
        _client = protocol == CollectorExporterProtocol.http
            ? _HTTPExporterClient(uri, httpClient ?? http.Client())
            : _GRPCExporterClient(pb_trace_service_grpc.TraceServiceClient(
                grpc.ClientChannel(
                  uri.host,
                  port: uri.port,
                ),
              ));

  @override
  void export(List<sdk.ReadOnlySpan> spans) {
    if (_isShutdown) {
      return;
    }

    if (spans.isEmpty) {
      return;
    }

    final body = pb_trace_service.ExportTraceServiceRequest(
        resourceSpans: _spansToProtobuf(spans));
    _client.send(body, _headers);
  }

...

  @override
  void shutdown() {
    _isShutdown = true;
    client.close();
    _client.close();
  }
}
```

huandu: I have also considered implementing this feature with this design, but ultimately decided against it because I felt that the design was overly complex and offered a low cost-performance ratio. As shown in the code you provided above, adopting this design would require us to implement several additional client classes, which are unlikely to be reused in the current project. Considering that these clients are only capable of sending trace information and are also unlikely to be reused in the future, the practical significance of such abstraction is not substantial. Moreover, this design does not completely eliminate the need for a switch statement. As can be seen in the `CollectorExporter`, there is a `? :` expression, which is essentially a switch statement, and it is not as clear as a switch statement. Therefore, I believe that this encapsulation does not completely eliminate switch statements but instead adds complexity.

To truly address the design issues caused by switch statements, ideally, we should use an Inversion of Control (IoC) approach to inject the clients implementing various protocols into the `CollectorExporter`, removing the `this.uri`, `httpClient`, and `protocol` parameters, and leaving the construction of the client to the caller of the library. This would best eliminate the switch statements. However, doing so would break backward compatibility with the original interface, which is a destructive upgrade. We certainly cannot make such a significant change to the interface.

Taking into account these considerations, I believe that the implementation approach in the current PR is more cost-effective than your proposal. Additionally, since the changes to the code in `CollectorExporter` are minimal and the logic is clearer, it would be more conducive to long-term maintenance in the future. Therefore, I hope you can reconsider this design proposal.

blakeroberts-wk: This makes sense. I agree with you.

It is worth note that the ternary expression in my example is done at instantiation whereas the switch statement in the current implementation is at invocation of export. With a batch processor, this creates marginal overhead. But a simple span processor might invoke export at a high frequency. Still, if this level of performance is of concern, Dart is probably not the appropriate language for the task.

I think that IoC would be an even better design. And this repository has maintained a pre-1.0 version to allow for backward incompatible changes. Of course we would need an intermediate version that maintained backward compatibility. This would most likely mean that the new implementation would need to exist as a new exporter class while the old exporter class remained. There will most likely be at least one renaming of this class prior to the 1.0 release of the software. Most likely, the renaming strategy would follow the JS implementation which uses several packages to accomplish IoC between protocol and platform: https://github.com/open-telemetry/opentelemetry-js/tree/main/experimental/packages.

Though I would argue supporting the feature, gRPC, in a backward compatible way with minimal changes and complexity, is valuable. So I'm on board.

.gitignore
```diff
@@ -1,3 +1,4 @@
 .dart_tool
 .packages
 pubspec.lock
+opentelemetry-proto
```
blakeroberts-wk, line 4: I am curious what workflow resulted in needing this change? Should there by an accompanying update to the Makefile that adds or fixes something in relation to contributing to the project?

lib/src/sdk/proto/opentelemetry/proto/collector/trace/v1/trace_service.pb.dart
```diff
@@ -1,6 +1,3 @@
-// Copyright 2021-2022 Workiva.
```
blakeroberts-wk, line 1: can you run `make format` to ensure the licensing notice comment appears at the top of all files?

.gitignore
huandu: I can rollback this change. I didn't realize that it was a submodule.

lib/src/sdk/proto/opentelemetry/proto/collector/trace/v1/trace_service.pb.dart
huandu: Sure.

huandu:
@blakeroberts-wk The CI job test-darv2 fails because that an outdated protobuf package (2.1.0) is installed but gRPC requires at least 3.x. I update the `pubspec.yaml` and set `protobuf: ^3.0.0`.

huandu:
@blakeroberts-wk From my observation, the `protoc_plugin` is the root cause of this issue. Starting from 21.0.0, it depends on `protobuf: ^3.0.0`. Unfortunately, the last gRPC package running on dart v2 depends on `protobuf: ^2.0.0`. I think one possible solution to fix this issue is to install `protoc_plugin` 20.0.1 on dart v2. It requires some hacks in `Makefile`.

robbecker-wf:
closing as stale

dustinlessard-wf:
Hi @huandu 
Thank you for taking the time to contribute to this project. We really appreciate the effort you put into opening this pull request.
I want to be transparent about the state of the repository: we have not been actively maintaining this codebase, and we do not currently have the bandwidth to properly review, test, and support external contributions at the standard they deserve.
For that reason, we’re going to close this pull request. This is not a reflection on the quality or value of your contribution. Rather, it reflects our current inability to responsibly vet and merge changes into the project.
We are also exploring a path to transfer ownership or stewardship of the project outside of our company, so that it can hopefully have a more active and sustainable future. Until that direction is clearer, we do not want to create a misleading impression that the repository is actively maintained or that we can support merged changes going forward.
Thank you again for your contribution and for your interest in improving the project. We’re sorry we were not able to engage with your pull request in the way it deserved.

PR #167, opened 2024-05-07
O11Y-4010: Add ContextManager for Context

changliu-wk:

- Add ContextManager interface
- Make Context abstract/interface.
    - The static current/root methods should be updated to pull from the static global ContextManager.
    - The default context manager should be the existing Dart zone implementation
    - Static context manager registration function is added to allow overriding context manager at runtime.
- Pull the Dart zone logic into its own implementation
- Experts necessary method to be public
Fixes # (issue)

## How Has This Been Tested?
CI test would be efficient. 

## Checklist:


aviary2-wf:
## Security Insights

## Action Items
---

lib/src/api/context/context.dart
```diff
@@ -31,14 +31,38 @@ import '../trace/nonrecording_span.dart';
...
+  static ContextManagerFunction? _contextManagerFunction;
+
+  static void RegisterContextManagerFunction(
+      ContextManagerFunction contextManagerFunction) {
+    _contextManagerFunction = contextManagerFunction;
+  }
+
   /// The active context.
-  static Context get current => Context._(Zone.current);
+  static Context get current => _getActive();
```
jonathancampbell-wk, line 52: In my opinion a ContextManager should be able to register an implementation of Context that overrides all Context functions. So for example if I did not want to use Dart Zones at all I could register a new class that implements all members of the Context class.

This does allow us to customize how we get active spans (only if there is no parent span in context) but it doesn't allow us to complete override the Dart context implementation.

blakeroberts-wk: It seems a little odd to start referencing a context manager that doesn't exist yet. Should the first PR add a context manager? Then, optionally, in the same PR or a follow up PR, this class can be updated to use the manager similar to how the js implementation handles it: https://github.com/open-telemetry/opentelemetry-js/blob/main/api/src/api/context.ts#L60

changliu-wk: It could be empty, I want to add one in wo11y-dart, something like this: 
```dart
api.SpanContext? GetSpanContextFromJSZone() {
  final jsContext = JsContextAPI.active();
  print('JsContextAPI.active(): $jsContext');
  final jsSpan = jsContext.getValue(JsSymbol.SymbolFromRegistry(spanKey.name));
  final traceId = jsSpan.spanContext().traceId;
  print('traceId: $traceId');
  final sc = api.SpanContext(
    api.TraceId.fromString(jsSpan.spanContext().traceId),
    api.SpanId.fromString(jsSpan.spanContext().spanId),
    jsSpan.spanContext().traceFlags,
    api.TraceState.empty()
  );
  return sc;
}
```

changliu-wk: In the example of js, the method is not `static`, in our case, the method is `static`, and in order to back compatible, we either do it like [this](https://github.com/Workiva/opentelemetry-dart/compare/master...jcampbell/context_api), or like what I did.

lib/src/api/context/context.dart
```diff
@@ -26,19 +26,28 @@
 /// OpenTelemetry SDKs.
```
jonathancampbell-wk, line 26: Should we remove the comment/docs about Zones since this is moved over to the implementation class?

lib/src/api/context/context.dart
```diff
@@ -26,19 +26,28 @@
...
-  final Zone _zone;
+/// The [ContextManager] is responsible for managing the current [Context].
+/// Different implementations of [ContextManager] can be registered to use different underlying storage mechanisms.
+/// The default implementation is [ZoneContextManager], which uses Dart zones to store the current [Context].
+ContextManager _contextManager = ZoneContextManager();
 
-  Context._(this._zone);
+abstract class Context {
+  static void RegisterContextManager(ContextManager contextManager) {
+    Logger('Context').info('Registering ContextManager: $contextManager');
```
jonathancampbell-wk, line 45: Could probably remove this log message. I added it for debugging purposes initially.

lib/src/api/context/context.dart
```diff
@@ -1,44 +1,25 @@
...
-/// (the rest of the values will be inherited from the forked Zone).
-///
-/// This library provides a simple abstraction over [Zone] for the purpose of
-/// implementing the rest of the Context specification. OpenTelemetry SDKs and
-/// instrumentation libraries should use this [Context] API instead of a [Zone]
-/// directly. Other users should usually not interact with Context at all and
-/// should instead manipulate it through cross-cutting concerns APIs provided by
-/// OpenTelemetry SDKs.
-import 'dart:async';
+import 'package:opentelemetry/src/sdk/context/zone_context_manager.dart';
```
blakeroberts-wk, line 3: The api should not depend on the sdk. one way to achieve this is to have the api implement a no-op context and context manager that is used as default

lib/src/api/context/context.dart
```diff
@@ -1,44 +1,25 @@
...
-class Context {
-  final Zone _zone;
+/// The [ContextManager] is responsible for managing the current [Context].
+/// Different implementations of [ContextManager] can be registered to use different underlying storage mechanisms.
+/// The default implementation is [ZoneContextManager], which uses Dart zones to store the current [Context].
+ContextManager _contextManager = ZoneContextManager();
 
-  Context._(this._zone);
+abstract class Context {
+  static void RegisterContextManager(ContextManager contextManager) {
```
blakeroberts-wk, line 17: I'm not sure if we should namespace this function as a method within the abstract `Context` class. Perhaps we pull it out and move it to the `context_manager.dart` file?

lib/src/api/context/context.dart
```diff
@@ -1,44 +1,25 @@
...
 
 /// [ContextKey] used to store spans in a [Context].
 final ContextKey spanKey = Context.createKey('OpenTelemetry Context Key SPAN');
 
-class Context {
-  final Zone _zone;
+/// The [ContextManager] is responsible for managing the current [Context].
+/// Different implementations of [ContextManager] can be registered to use different underlying storage mechanisms.
+/// The default implementation is [ZoneContextManager], which uses Dart zones to store the current [Context].
+ContextManager _contextManager = ZoneContextManager();
```
blakeroberts-wk, line 14: I feel like this would be more appropriately located in the `context_manager.dart` file. And in that case, the static methods of `Context` should be removed or updated to use a global accessor function

changliu-wk: If doing this, we need to get active context by doing, api.ContextManager.current instead of api.Context.current?

lib/src/sdk/context/zone_context_manager.dart
```diff
@@ -0,0 +1,13 @@
+// Copyright 2021-2022 Workiva.
+// Licensed under the Apache License, Version 2.0. Please see https://github.com/Workiva/opentelemetry-dart/blob/master/LICENSE for more information
+
+import '../../../api.dart';
+import 'zone_context.dart';
+
+class ZoneContextManager implements ContextManager {
+  @override
+  Context get activeContext => ZoneContext.current;
```
blakeroberts-wk, line 9: Implementing context manager seems unnecessary. Perhaps remove abstract from context manager and remove this class

test/integration/sdk/context_test.dart
```diff
@@ -9,17 +9,17 @@ void main() {
   test('setValue and getValue', () {
```
blakeroberts-wk, line 9: Arbitrary line: should this test remain in api or be renamed to zone context test?

lib/src/api/context/context_manager.dart
```diff
@@ -0,0 +1,23 @@
+// Copyright 2021-2022 Workiva.
+// Licensed under the Apache License, Version 2.0. Please see https://github.com/Workiva/opentelemetry-dart/blob/master/LICENSE for more information
+
+import 'package:opentelemetry/api.dart';
+import 'package:opentelemetry/src/sdk/context/zone_context_manager.dart';
+
+abstract class ContextManager {
```
blakeroberts-wk, line 7: Can the class be marked static? Or perhaps its methods should become top level functions within this package?

lib/src/api/context/context.dart
```diff
@@ -1,54 +1,8 @@
...
-/// [ContextKey] used to store spans in a [Context].
-final ContextKey spanKey = Context.createKey('OpenTelemetry Context Key SPAN');
-
-class Context {
-  final Zone _zone;
-
-  Context._(this._zone);
-
-  /// The active context.
-  static Context get current => Context._(Zone.current);
```
jonathancampbell-wk, line 41: Does removing `current` and `root` create a breaking change?

changliu-wk: Yes, it will.

changliu-wk:
After discussed with @blakeroberts-wk, we decided to revert the breaking changes, we will have follow up tickets about the API redesign.

lib/src/api/context/context.dart
changliu-wk: I reverted it, it won't  have breaking changes.

jonathancampbell-wk:
> After discussed with @blakeroberts-wk, we decided to revert the breaking changes, we will have follow up tickets about the API redesign.

Which changes do we want to push to future PRs?

changliu-wk:
> > After discussed with @blakeroberts-wk, we decided to revert the breaking changes, we will have follow up tickets about the API redesign.
> 
> Which changes do we want to push to future PRs?



> > After discussed with @blakeroberts-wk, we decided to revert the breaking changes, we will have follow up tickets about the API redesign.
> 
> Which changes do we want to push to future PRs?

I think we would like to mimic what opentelemetry-js does, with some adjustments, like not having delete method within context api.

lib/src/api/context/context.dart
jonathancampbell-wk: Or move the default dart zone implementation into the API?

lib/src/api/context/context_manager.dart
blakeroberts-wk: talked in person, current implementation is sufficient

lib/src/api/context/context_manager.dart
```diff
@@ -0,0 +1,10 @@
+// Copyright 2021-2022 Workiva.
+// Licensed under the Apache License, Version 2.0. Please see https://github.com/Workiva/opentelemetry-dart/blob/master/LICENSE for more information
+
+import 'package:opentelemetry/api.dart';
+
+abstract class ContextManager {
+  Context get root;
```
blakeroberts-wk, line 7: can we mark this as deprecated since we expect it to be removed in a future version?

lib/src/api/context/context.dart
```diff
@@ -47,7 +28,7 @@ class Context {
   /// Only use this context if you are certain you need to disregard the
   /// current [Context].  For example, when instrumenting an asynchronous
   /// event handler which may fire while an unrelated [Context] is "current".
-  static Context get root => Context._(Zone.root);
+  static Context get root => _contextManager.root;
```
blakeroberts-wk, line 31: can we mark this as deprecated?

lib/src/api/context/context.dart
blakeroberts-wk: can we mark this method as experimental and move it to a top level function?

lib/src/sdk/context/zone_context_manager.dart
```diff
@@ -0,0 +1,13 @@
...
+
+import '../../../api.dart';
+import 'zone_context.dart';
+
+class ZoneContextManager implements ContextManager {
+  @override
+  Context get active => ZoneContext.current;
+
+  @override
+  Context get root => ZoneContext.root;
```
blakeroberts-wk, line 12: can we mark this as deprecated?

lib/src/sdk/context/zone_context_manager.dart
```diff
@@ -0,0 +1,15 @@
+// Copyright 2021-2022 Workiva.
+// Licensed under the Apache License, Version 2.0. Please see https://github.com/Workiva/opentelemetry-dart/blob/master/LICENSE for more information
+
+import '../../../api.dart';
+import 'zone_context.dart';
+
+class ZoneContextManager implements ContextManager {
```
blakeroberts-wk, line 7: perhaps we also add the experimental annotation to this class as well? https://pub.dev/documentation/meta/latest/meta/experimental-constant.html

lib/src/api/context/context.dart
```diff
@@ -1,44 +1,25 @@
...
 
-  Context._(this._zone);
+abstract class Context {
+  static void RegisterContextManager(ContextManager contextManager) {
+    _contextManager = contextManager;
+  }
 
   /// The active context.
-  static Context get current => Context._(Zone.current);
+  static Context get current => _contextManager.active;
```
blakeroberts-wk, line 22: should we deprecate this in favor of a top level function?

changliu-wk: We don't have a schedule and priority for this work, I hope it won't be overwhelming for the user?

lib/src/api/context/context.dart
blakeroberts-wk: I think we should default to use a map implementation like the js implementation. but if we would not like making that change in this PR, I think it makes sense to move the dart zone implementation to the api

lib/src/api/context/context.dart
blakeroberts-wk: we could define a package private getter that the static Context method can use

lib/src/api/context/context.dart
blakeroberts-wk: we can hold on this decision for now

lib/src/api/context/context.dart
jonathancampbell-wk: I think it's a reasonable change for this PR. My one concern is introducing a breaking change to functionality, would users lose context propagation across async code if they updated to the latest version?

blakeroberts-wk: yeah. we'd probably want to indicate this somehow. maybe release notes would be sufficient but we could add a deprecation annotation too that simply warns that the default functionality will change

lib/src/api/context/map_context.dart
```diff
@@ -0,0 +1,46 @@
+// Copyright 2021-2022 Workiva.
+// Licensed under the Apache License, Version 2.0. Please see https://github.com/Workiva/opentelemetry-dart/blob/master/LICENSE for more information
+
+import '../../../api.dart';
+import 'context.dart';
+
+class MapContext implements Context {
+  final Map<ContextKey, Object> _contextMap = {};
```
changliu-wk, line 8: static map

lib/src/api/context/noop_context_manager.dart
```diff
@@ -0,0 +1,30 @@
+// Copyright 2021-2022 Workiva.
+// Licensed under the Apache License, Version 2.0. Please see https://github.com/Workiva/opentelemetry-dart/blob/master/LICENSE for more information
+
+import 'context.dart';
+import 'context_manager.dart';
+import 'map_context.dart';
+
+class NoopContextManager implements ContextManager {
+  final MapContext _rootContext = MapContext();
```
changliu-wk, line 9: Move it out

lib/src/api/trace/nonrecording_span.dart
```diff
@@ -10,8 +10,6 @@ import '../../../api.dart' as api;
 /// See https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#wrapping-a-spancontext-in-a-span
 /// for more information.
 ///
-/// This class should not be exposed to consumers and is used internally to wrap
-/// [api.SpanContext] being injected or extracted for external calls.
```
changliu-wk, line 14: Check it

lib/api.dart
```diff
@@ -4,7 +4,9 @@
 export 'src/api/common/attribute.dart' show Attribute;
 export 'src/api/common/resource_attributes.dart' show ResourceAttributes;
 export 'src/api/common/semantic_attributes.dart' show SemanticAttributes;
-export 'src/api/context/context.dart' show Context;
+export 'src/api/context/context.dart' show Context, ContextKey;
+export 'src/api/context/noop_context_manager.dart'
+    show registerContextManager, getContextManager;
```
changliu-wk, line 9: move to experimental_api

lib/api.dart
```diff
@@ -22,6 +24,7 @@ export 'src/api/propagation/w3c_trace_context_propagator.dart'
     show W3CTraceContextPropagator;
 export 'src/api/span_processors/span_processor.dart' show SpanProcessor;
 export 'src/api/trace/id_generator.dart' show IdGenerator;
+export 'src/api/trace/nonrecording_span.dart' show NonRecordingSpan;
```
changliu-wk, line 27: move to experimental as well

lib/src/api/context/context.dart
```diff
@@ -58,33 +31,36 @@ class Context {
...
   /// [api.SpanContext] exists.
-  api.SpanContext get spanContext =>
-      (getValue(spanKey) ?? NonRecordingSpan(api.SpanContext.invalid()))
-          .spanContext;
+  @Deprecated(
+      'This method will be removed in the future, new method will be added.')
+  api.SpanContext get spanContext;
 }
 
 class ContextKey {
```
changliu-wk, line 66: typedef ContextKey = String; https://dart.dev/language/typedefs

lib/src/api/context/context.dart
```diff
@@ -1,44 +1,17 @@
...
 
-class Context {
-  final Zone _zone;
-
-  Context._(this._zone);
-
+abstract class Context {
   /// The active context.
-  static Context get current => Context._(Zone.current);
+  static Context get current => getContextManager().active;
```
blakeroberts-wk, line 14: I think we can mark this as deprecated

blakeroberts-wk: Either we can make it a top level function, or just let consumers grab it from the registered context manager directly

lib/src/api/context/noop_context_manager.dart
```diff
@@ -0,0 +1,32 @@
...
+/// in the future, this will be replaced with map context manager.
+ContextManager _contextManager = ZoneContextManager();
+
+void registerContextManager(ContextManager contextManager) {
+  _contextManager = contextManager;
+}
+
+ContextManager getContextManager() {
+  return _contextManager;
+}
```
blakeroberts-wk, line 32: move these to `context_manager.dart` and follow a pattern similar to the other globals: https://github.com/Workiva/opentelemetry-dart/blob/master/lib/src/api/open_telemetry.dart#L11-L28

lib/src/api/context/context.dart
```diff
@@ -47,50 +20,47 @@ class Context {
...
-}
-
-class ContextKey {
-  /// Name of the context key.
-  final String name;
-
-  /// Construct a [ContextKey] with a given [name].
-  ContextKey(this.name);
+  @Deprecated(
+      'This method will be removed in the future, new method will be added.')
```
blakeroberts-wk, line 64: I don't think we're going to add a new method for this

jonathancampbell-wk: Are we sure we don't want it? It's outside of the spec but is a useful utility to have. [Here's the JS example](https://github.com/open-telemetry/opentelemetry-js/blob/e49c4c7f42c6c444da3f802687cfa4f2d6983f46/api/src/api/trace.ts#L108)

blakeroberts-wk: I'm not sure how useful it is. I think `span.spanContext` is adequate

lib/src/api/context/context.dart
changliu-wk: Do we want to move it now or later?

blakeroberts-wk: we can't move it now or that'd break api, but we can deprecate for now and decide if we want the top level api or not later

lib/src/api/context/context.dart
jonathancampbell-wk: Fair enough, it should be simple enough to access a spanContext assuming getSpan is available.

blakeroberts-wk: yes. which we'll want to introduce top level api to create something like `final spanCtx = fromContext(ctx)?.spanContext`

changliu-wk: That was what I meant for the `new method`, since it is from span api, which we already have, I will remove it.

lib/src/api/context/context.dart
```diff
@@ -47,50 +20,45 @@ class Context {
...
+  @Deprecated('We are planning to remove this in the future.')
+  static Context get root => globalContextManager.root;
 
   /// Returns a key to be used to read and/or write values to a context.
   ///
   /// [name] is for debug purposes only and does not uniquely identify the key.
   /// Multiple calls to this function with the same [name] will not return
   /// identical keys.
-  static ContextKey createKey(String name) => ContextKey(name);
+  static ContextKey createKey(String name) => name;
```
blakeroberts-wk, line 31: I think we can deprecate this in favor of something like `ContextKey(name)`?

lib/src/api/context/context_manager.dart
```diff
@@ -0,0 +1,32 @@
...
+
+import 'package:opentelemetry/api.dart';
+
+import 'zone_context_manager.dart';
+
+/// The [ContextManager] is responsible for managing the current [Context].
+/// Different implementations of [ContextManager] can be registered to use different underlying storage mechanisms.
+abstract class ContextManager {
+  @Deprecated(
+      'We are planning to remove this in the future, please use Context.current instead.')
```
blakeroberts-wk, line 12: `Context.current` will be going away too. They'll need to use `contextManager.active`

lib/src/api/context/context_manager.dart
```diff
@@ -0,0 +1,32 @@
...
+/// in the future, this will be replaced with noop context manager which maintains map context.
+final ContextManager _noopcontextManager = ZoneContextManager();
+ContextManager _contextManager = _noopcontextManager;
+
+ContextManager get globalContextManager => _contextManager;
+
+void registerGlobalContextManager(ContextManager contextManager) {
+  if (_contextManager != _noopcontextManager) {
+    throw StateError(
+        'Global context manager is already registered, registerContextManager must be called only once before any calls to the getter globalContextManager.');
```
blakeroberts-wk, line 28: The last part about needing to call this before the getter seems out of place

changliu-wk: Since there is a default context manager, so we don't need to mention this must be called before this getter, I think.

lib/src/api/context/context.dart
```diff
@@ -1,44 +1,14 @@
...
-
-/// [ContextKey] used to store spans in a [Context].
-final ContextKey spanKey = Context.createKey('OpenTelemetry Context Key SPAN');
-
-class Context {
-  final Zone _zone;
+import 'context_manager.dart';
 
-  Context._(this._zone);
+typedef ContextKey = String;
```
blakeroberts-wk, line 6: we actually shouldn't type alias string for the context key. The reason is that this allows access to context values using the string value used as the key. The intent of type aliasing is to prohibit retrieval of the value without the key used to store the value. That is why a class is used. The class instance is used as the index rather than the string value.

For example, with the current type alias version, someone can call `ctx.getValue('OpenTelemetry Context Key SPAN')` and this will return the span from the context. But this allows for key conflicts within the context. The value of the span should only be accessible via `ctx.getValue(spanKey)`.

Copy/paste the following into https://dartpad.dev/ for an example:

```dart
typedef ContextKey = String;

final ContextKey _spanKey = "private key for accessing private value of a map";

class ContextKeyClass {
  final String _name;
  ContextKeyClass(this._name);
}

final ContextKeyClass _spanKeyClass = ContextKeyClass("actually private");

void main() {
  final Map<ContextKey, dynamic> m = Map();
  m[_spanKey] = "value that should only be accessed using [_spanKey]";
  print(m[
      "private key for accessing private value of a map"]); // should print null but prints value

  final Map<ContextKeyClass, dynamic> m2 = Map();
  m2[_spanKeyClass] =
      "value that should only be accessed using [_spanKeyClass]";
  print(m2[ContextKeyClass("actually private")]); // should print null and does
  print(m2[_spanKeyClass]); // should print value and does
}
```

The use of the `_name` field is for demonstration purposes only. The actual implementation of the class can remove this:

```dart
class ContextKey {}

void main() {
  final Map<ContextKey, dynamic> m = Map();
  m[spanKey] = "value that should only be accessed using [spanKey]";
  print(m[spanKey]);
  print(m[ContextKey()]);
}

```

blakeroberts-wk (approved):

changliu-wk:
@Workiva/release-management-p

changliu-wk:
QA+1

rmconsole-wf (approved):
+1 from RM

Issue #174, opened 2024-06-09
Usage examples for async would be really valuable


time4tea:
Hiya - Thanks for this library! 

I'm trying to instrument an async process, and its not clear to me how to track the call through the system.

For example, plugging into 'dio' using interceptors, i can create a new span like this:

```dart 

  dynamic startHttpRequest(RequestOptions options, handler) {
    var span =
        tracer.startSpan("http", context: ZoneContext.current, attributes: [
      Attribute.fromString(SemanticAttributes.httpMethod, options.method),
      Attribute.fromString(SemanticAttributes.httpHost, options.uri.host),
      Attribute.fromString(SemanticAttributes.httpScheme, options.uri.scheme),
    ]);

    print(span.spanContext.spanId);

    ZoneContext.current.withSpan(span).execute(() {
      handler.next(options);
    });
  }

```

and then I'm hoping to be able to close the span at the end of the request like this:

```dart
  dynamic endHttpRequest(Response<dynamic> response, handler) {
    var span = ZoneContext.current.span;

    print(span.spanContext.spanId);

    span.setAttributes([
      Attribute.fromInt(SemanticAttributes.httpStatusCode, response.statusCode!)
    ]);

    span.end();
    handler.next(response);
  }

```

however, the spanid at the end is the invalid span, so i guess I'm not doing the right thing with the context. Are there any examples for how to instrument something like `dio`?

Of course - if i get this working, it could be contributed, if that were useful.

Thank you!

ukasz123:
@time4tea Following interceptor implementation worked for me.

```dart
import 'package:dio/dio.dart';
import 'package:opentelemetry/api.dart' as otel_api;

class TracingInterceptor implements Interceptor {
  static const _spanExtra = 'telemetry_span';

  static const _tracerName = 'HTTP-client';

  List<otel_api.Attribute> _getRequestAttributes(RequestOptions options) {
    // NO-OP
    return [];
  }

  String _getSpanName(RequestOptions options) {
    // NO-OP
    return 'http';
  }

  @override
  void onRequest(RequestOptions options, RequestInterceptorHandler handler) {
    final requestSpan =
        otel_api.globalTracerProvider.getTracer(_tracerName).startSpan(
              _getSpanName(options),
              kind: otel_api.SpanKind.client,
              attributes: _getRequestAttributes(options),
            );
    // store the span object for later
    options.extra[_spanExtra] = requestSpan;

    handler.next(options);
  }

  @override
[...]

PR #176, opened 2024-06-19
add cross-cutting context api functions

### LINKED ISSUE #173: [Question] What is correct way to call Context.current.span now?
Prior to 0.18.1 I had used `Context.current.span` to access the current span and add attributes to it in my code.  

This approach has been deprecated in recent versions, but the deprecation notes don't suggest or point to what to use instead.  

I'm struggling to find documentation to tell me what the recommended way to do this is now.  Any pointers? 🙏🏼

blakeroberts-wk:

The context API has deprecated members without appropriate replacements.

Fixes #173, #174

The following functions were added to the API:

- `contextWithSpan`,
- `contextWithSpanContext`
- `spanContextFromContext`
- `spanFromContext`
- `traceContext ` (experimental)
- `traceSyncContext ` (experimental)

The following functions from the API were deprecated:

- `trace`
- `traceSync`

The following changes were made to implementations but not interfaces to avoid breaking changes:
- `api.NoopTracer.start()` and `sdk.Tracer.start()` support an optional argument `bool newRoot`. The API's `Tracer` interface will include this option in the 0.19.0 release.
- `api.ZoneContext.run()` and `api.MapContext.run()` methods were added. The API's `Context` interface will include this method in the 0.19.0 release.

## How Has This Been Tested?

The three new examples added produce expected output.

## Checklist:


aviary-wf:
## Security Insights

## Action Items
---

blakeroberts-wk:
QA +1 changes were tested on a dart browser app with both zone and map context propagation. Examples succinctly demonstrate functionality as well

lib/src/api/context/context.dart
```diff
@@ -17,16 +38,18 @@ abstract class Context {
   /// Only use this context if you are certain you need to disregard the
   /// current [Context].  For example, when instrumenting an asynchronous
   /// event handler which may fire while an unrelated [Context] is "current".
-  @Deprecated('We are planning to remove this in the future.')
-  static Context get root => globalContextManager.root;
+  @Deprecated(
+      'This method will be removed in 0.19.0. Use [contextWithSpanContext(globalContextManager.active, SpanContext.invalid())] instead.')
+  static Context get root => contextWithSpanContext(
+      globalContextManager.active, SpanContext.invalid());
```
changliu-wk, line 44: It took me a while to understand how it works, is that possible to simplify this effort and put it as a mehod for globalContextManager or ContextManager?

blakeroberts-wk: The purpose of this api is to allow creating new root spans. Rather than having users reference a root context when creating a span, `tracer.startSpan()` will include an optional named parameter `bool newRoot`. Because adding that parameter to the tracer interface is a breaking change, it will need to wait until v0.19.0. For now, folks could cast their api tracer to an sdk tracer and use `newRoot`. Or, as the deprecated annotation suggests, manually construct a context with an invalid span context. Either way, these are short term work arounds until a more succinct solution can be introduced

lib/src/sdk/trace/tracer.dart
```diff
@@ -35,35 +36,34 @@ class Tracer implements api.Tracer {
       api.SpanKind kind = api.SpanKind.internal,
       List<api.Attribute> attributes = const [],
       List<api.SpanLink> links = const [],
-      Int64? startTime}) {
-    context ??= api.Context.current;
+      Int64? startTime,
+      bool newRoot = false}) {
```
changliu-wk, line 40: Add a test case for this new parameter?

lib/src/api/context/map_context.dart
```diff
@@ -1,14 +1,16 @@
...
+import 'package:meta/meta.dart';
 
-final ContextKey spanKey = ContextKey();
+import '../../../api.dart';
 
 class MapContext implements Context {
   final Map<ContextKey, Object> _contextMap = {};
 
+  @protected
+  MapContext();
```
evanweible-wf, line 12: The [`@protected` annotation](https://pub.dev/documentation/meta/latest/meta/protected-constant.html) is typically used on instance members and not constructors. Is the goal to prevent consumers from directly constructing instances of `MapContext`? If so, the idiomatic way to do that in Dart is to define a private constructor and a separate factory function that is not publicly exported:

```dart
// not publicly exported
MapContext createMapContext() => MapContext._();

// publicly exported
class MapContext implements Context {
  MapContext._();
  ...
}
```

On this topic - once we drop support for Dart 2, we'll be able to leverage the new class modifiers that Dart 3 supports:
- https://dart.dev/language/class-modifiers
- https://dart.dev/language/class-modifiers-for-apis

lib/src/api/context/map_context.dart
```diff
@@ -30,7 +32,7 @@ class MapContext implements Context {
   /// Returns a new [MapContext] created from this one with the given [Span]
   /// set.
   @override
-  MapContext withSpan(Span span) => setValue(spanKey, span);
+  Context withSpan(Span span) => contextWithSpan(this, span);
```
evanweible-wf, line 35: fyi this is technically a breaking change since `Context` is less specific than `MapContext`. Would only affect consumers if they were typing the return value from a call to `withSpan` as `MapContext` or if they were using an API that exists only on `MapContext` and not `Context` (which I don't think is an issue since there are currently no APIs unique to this subclass.

lib/src/api/context/map_context.dart
```diff
@@ -39,10 +41,14 @@ class MapContext implements Context {
   /// Get the [Span] attached to this [MapContext], or an invalid, [Span] if no such
   /// [Span] exists.
   @override
-  Span get span => getValue(spanKey) ?? NonRecordingSpan(SpanContext.invalid());
+  Span get span => spanFromContext(this);
+
+  /// Call [fn] with this [MapContext] and return its result.
+  @experimental
+  R run<R>(R Function(Context context) fn) => fn(this);
```
evanweible-wf, line 48: What's the use case for this? Seems to me like it would be simpler for the caller to just run `fn(this)` themselves since they'd have to have a reference to the function and the context in order to call this method.

lib/src/api/context/zone_context.dart
```diff
@@ -26,28 +26,15 @@
...
-  /// The root context which all other contexts are derived from.
-  ///
-  /// It should generally not be required to use the root [ZoneContext] directly -
-  /// instead, use [ZoneContext.current] to operate on the current [ZoneContext].
-  /// Only use this context if you are certain you need to disregard the
-  /// current [ZoneContext].  For example, when instrumenting an asynchronous
-  /// event handler which may fire while an unrelated [ZoneContext] is "current".
-  static ZoneContext get root => ZoneContext._(Zone.root);
+  @protected
+  ZoneContext(this._zone);
```
evanweible-wf, line 37: Same question/note about the protected constructor here.

lib/src/api/context/zone_context.dart
```diff
@@ -60,25 +47,29 @@ class ZoneContext implements Context {
...
-  ZoneContext setValue(ContextKey key, Object value) =>
-      ZoneContext._(_zone.fork(zoneValues: {key: value}));
+  Context setValue(ContextKey key, Object value) =>
+      ZoneContext(_zone.fork(zoneValues: {key: value}));
 
   /// Returns a new [ZoneContext] created from this one with the given [Span]
   /// set.
   @override
-  ZoneContext withSpan(Span span) => setValue(spanKey, span);
+  Context withSpan(Span span) => contextWithSpan(this, span);
```
evanweible-wf, line 56: And same not here about the return type being less specific which is technically a breaking change. Unlikely that it would impact anyone, but wanted to note it.

lib/src/api/context/zone_context.dart
```diff
@@ -60,25 +47,29 @@ class ZoneContext implements Context {
...
 
   /// Execute a function [fn] within this [ZoneContext] and return its result.
   @override
   R execute<R>(R Function() fn) => _zone.run(() => fn());
 
+  /// Call [fn] in this [ZoneContext]'s [Zone] and return its result.
+  @experimental
+  R run<R>(R Function(Context context) fn) => _zone.run(() => fn(this));
```
evanweible-wf, line 64: Now that I'm seeing this here as well, should `run` be a part of the `Context` interface? Or are we holding off on that because it's experimental?

lib/src/api/open_telemetry.dart
```diff
@@ -39,15 +42,52 @@ void registerGlobalTextMapPropagator(api.TextMapPropagator textMapPropagator) {
...
+    rethrow;
+  } finally {
+    span.end();
+  }
+}
 
+/// Records a span of the given [name] for the given function with a given
+/// [api.Tracer] and marks the span as errored if an exception occurs.
+@experimental
+Future<T> traceContext<T>(String name, Future<T> Function(api.Context) fn,
```
evanweible-wf, line 68: Okay, I think I see now why having `run` defined on the `Context` classes is useful now. Which makes me think it should be defined on the parent `Context` class, too.

lib/src/api/open_telemetry.dart
```diff
@@ -39,15 +42,52 @@ void registerGlobalTextMapPropagator(api.TextMapPropagator textMapPropagator) {
...
+  if (newRoot) {
+    context = api.contextWithSpanContext(context, api.SpanContext.invalid());
+  }
+
+  final span = tracer.startSpan(name,
+      context: context, kind: spanKind, links: spanLinks);
+  context = api.contextWithSpan(context, span);
   try {
-    return await context.withSpan(span).execute(fn);
+    // TODO: remove this check once `run` exists on context interface
```
evanweible-wf, line 86: I think this answers my question - we're avoiding adding it to the interface right now because it would be breaking to any implementers?

lib/src/api/context/map_context.dart
blakeroberts-wk: Nice catch. I may have made this mistake elsewhere 🤔 but at least for `MapContext`, it hasn't been made public yet. Any reference would be from something inside `src` such as `src/experimental_api.dart`

lib/src/api/context/map_context.dart
blakeroberts-wk: yes. I'll update this.

lib/src/api/context/zone_context.dart
blakeroberts-wk: same response: needs updated 😄

lib/src/api/context/map_context.dart
blakeroberts-wk: yes. I'm still not sold on this yet, but right now it's for api completeness because our ZoneContext impl needs a way to execute a function within its private zone reference.

I'd like to come up with a paradigm so the Context api doesn't need a run method. However, this requires the consumer to have and/or use a reference to a zone. Which is almost okay except for libraries. By having consumers need to reference the zone, libraries have to worry about portability of their tracing code vs just calling the run function whether it's needed or not (whether their consumers have registered a zone context manager or are using the no op context manager)

lib/src/api/context/zone_context.dart
blakeroberts-wk: Holding off because that'd be a breaking change. My current plan is to add the run method to the context api in the 0.19.0 release and still mark it as experimental at least for the time being 😅 I don't like it on the context api but I don't know how else to support zones and non-zones seamlessly

lib/src/api/open_telemetry.dart
blakeroberts-wk: Yes. Lol I should have read your entire review before responding 😆

lib/src/api/context/zone_context.dart
blakeroberts-wk: Same as other, this hasn't been exposed yet. Any reference to the `ZoneContext` type would be from `src`

blakeroberts-wk:
QA +1

test/integration/open_telemetry_test.dart
```diff
@@ -22,8 +22,8 @@ void main() {
         sdk.SpanLimits());
     late Span span;
 
-    api.traceSync('syncTrace', () {
-      span = api.Context.current.span as Span;
+    api.traceContextSync('syncTrace', (context) {
+      span = api.spanFromContext(context) as Span;
```
evanweible-wf, line 26: Are these `as Span` casts still necessary? Looks like `spanFromContext` should be typed to return `Span`

evanweible-wf (approved):

test/integration/open_telemetry_test.dart
blakeroberts-wk: Yes. The cast is to an internal `Span` type from the SDK, not the API. The SDK type is not publicly exposed

blakeroberts-wk:
QA +1 CI passes, examples produce expected output, and consumed the changes into a private project that was able to accomplish context propagation via zones without using non-exported types:

```dart
  final tracer = otel.globalTracerProvider.getTracer('my-tracer');
  final span = tracer.startSpan('my-span');
  final ctx = otel.contextWithSpan(otel.globalContextManager.active, span);
  otel.spanFromContext(ctx).end();
```

blakeroberts-wk:
@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #192, opened 2024-10-14
PPI-203 : deprecate context manager and implement attach/detach APIs

### LINKED ISSUE #185: Trace Context API Consolidation
### Is your feature request related to a problem?

The current `Context` implementation is as an interface where a registered `ContextManager` supplies the implementation. This allows a consumer (such as a library) to rely on the `Context` interface without needing to worry about the concrete implementation. In other words, the consumer does not need to worry about whether the `Context` implementation is a `ZoneContext` or `MapContext`.

However, in order to propagate context using zones, code must execute from the scope of `Zone.run()`. But to manually propagate context using a `MapContext`, no such requirement exists. As a result, the `Context` API has to supply a `run()` method to support one of the two context propagation methods.

Now, let's take a closer look at the two supported modes of context propagation: zone context propagation and map context propagation. Zone context propagation uses zones, simple enough. Map context propagation isn't really a unique propagation strategy. It is manual propagation. The Context stores its values in a map, but this has nothing to do with propagation. Rather than using a map, a zone can hold those values too. That `ZoneContext` can be propagated manually just the same as a `MapContext`.

So if a `ZoneContext` can do anything a `MapContext` can do, _and_ a `ZoneContext` can propagate context automatically too, what's the point of having a `MapContext`?

1. Language Support - If zones were not a native language feature (like with JavaScript), using zones would require a dependency on a third party library.
2. Performance - If forking zones was slow or memory intensive.
3. Unknown considerations.

(1) isn't the case since Dart supports zones natively. I haven't seen evidence to support (2). All that remains are unknowns, (3). I don't think it's wise to code against unknown considerations. To code for extensibility is good. But again, "anything [a map] can do [a zone] can do better 🎵."

### Describe the solution you'd like

Simple, intuitive APIs.

### Describe alternatives you've considered

```dart
final ContextKey contextKey = ContextKey();

Zone zoneWithContext(Context context) {
	return Zone.current
		.fork(zoneValues: {contextKey: context});
}

Context contextFromZone({Zone? zone}) {
	return (zone ?? Zone.current)[contextKey] as Context;
}

Span activeSpan() {
  return spanFromContext(contextFromZone());
}

Context rootContext() {
  return contextFromZone(zone: Zone.root);
}

void main() {
  final tracer = globalTracerProvider.getTracer('tracer');
  final parent = tracer.startSpan('parent')..end();
  final context = contextWithSpan(rootContext(), parent);

  zoneWithContext(context).run(() {
    // active returns the parent span
    activeSpan().addEvent('Hello, World!');

    // starting a span within this scope will automatically set its parent
    tracer.startSpan('child - automatic').end();
  });

  // manually propagating context works too
  tracer.startSpan('child - manual', context: context).end();
}
```

The above shows cross-cutting API additions that allow a `Context` to propagate via a `Zone`. This would allow `Context` to be a single concrete implementation.

We could take this even further and remove `Context` altogether in favor of using `Zone` directly in all cases. This would be achieved by replacing the current cross-cutting APIs for `Context` with `Zone` specifics:

```dart

Zone zoneWithSpan(Span span) {
  return Zone.current.fork(zoneValues: {
    spanKey: span,
  });
}

Zone zoneWithSpanContext(SpanContext spanContext) {
  return Zone.current.fork(zoneValues: {
    spanKey: NonRecordingSpan(spanContext),
  });
}

Span spanFromZone({Zone? zone}) {
  return (zone ?? Zone.current)[spanKey];
}

SpanContext spanContextFromZone({Zone? zone}) {
  return spanFromZone(zone: zone).spanContext;
}
```

The code snippet above:
- Is functional equivalent to what is supplied by all files within [src/api/context](https://github.com/Workiva/opentelemetry-dart/tree/008b9451d013541587fa3e9bfd94e855c4a302b2/lib/src/api/context). I.e. those files can be replaced by this snippet of code, alone.
- The snippet of code is in line with the [specification](https://opentelemetry.io/docs/specs/otel/context/) in that a `Zone` is immutable and capable of getting and setting values.

Here are some example usages of this set of APIs:

```dart
void main() {
  final tracer = globalTracerProvider.getTracer('tracer');
  final parent = tracer.startSpan('parent')..end();
  final zone = zoneWithSpan(parent)
    ..run(() {
      // by default, `spanFromZone()` returns the zone's "active" span
      spanFromZone().addEvent('Hello, World!');

      // starting a span within this scope will automatically set its parent
      tracer.startSpan('child - automatic').end();
    });

  // manual propagation would work too
  tracer.startSpan('child - manual', zone: zone).end();

  // The span of a zone not in scope can be retrieved too.
  final otherSpan = spanFromZone(zone: otherZone);

  // Using the root zone will always imply an invalid non-recording span.
  // As consequence, a new root span can be started by referencing the root zone:
  tracer.startSpan('new root', zone: Zone.root).end();
  // this could be simplified by adding a `bool newRoot` parameter:
  tracer.startSpan('new root 2', newRoot: true).end();
  // this parameter would have less value compared to the current implementation:
  tracer.startSpan('new root 3',
      context: contextWithSpanContext(
          globalContextManager.active, SpanContext.invalid()));
}
```

### Additional context

https://github.com/Workiva/opentelemetry-dart/pull/176

blakeroberts-wk:

Fixes #185 

The current context API is unnecessarily complex and does not support the attach/detach optional APIs.

In async programming, the attach/detach APIs can be incredibly handy in passing context to async processes that cannot propagate context manually (most likely due to public or otherwise unbreakable APIs). Zones can help in this case, but do not always suffice since zones might not appropriately follow code execution. Specifically, if a process includes putting an event on a stream/sink where that event is consumed by a stream listener, the zone does not propagate from where the event was placed on the stream/sink to the stream listener's callback.

The current context implementation requires a modifiable concrete implementation used to decide how to propagate context. This is an adaption to OpenTelemetry JavaScript's implementation.

However, unlike JavaScript, zones are a language native feature: its type and implementation provided by the standard library. Therefore it is okay to assume that zones are available and thus simplify the context propagation story as compared to OpenTelemetry JavaScript.

A potential solution to this end would be to replace the abstract `Context` with the concrete `ZoneContext`. This solution was avoided because `ZoneContext` presumes that zone values are what carries the context. This means that manual propagation would be passing around the zone itself (or rather the contrived wrapper of the zone).

Instead, the changes make `Context` its own immutable propagation context. This allows a zone, also an immutable propagation context, to hold both a stack of versions (due to multiple calls to attach) or a specific version (when binding a version to a zone for zone execution).

tl;dr; `Context` is a stack implemented as a linked list. Cross-cutting APIs allow stacking the stack (attach/detach) on a zone, or binding a specific version (linked list node) to a zone.

`Zone zoneWithContext(Context context)` -> binds a specific version to a fork of the current zone

`ContextToken attach(Context context)` -> stacks a specific version to the current zone

With the new APIs, retrieving the active context follows a hierarchy:
1. active attached context
2. active zone context
3. root context

Starting a span is the same: a manually specified context is used if given and the active context is used otherwise. The changed part is that the active context might be an attached context.

## How Has This Been Tested?

The provided example and unit tests demonstrates functionality. Additional testing will be carried out by consuming the changes in various test applications.

## Checklist:


aviary-wf:
## Security Insights

## Action Items
---

example/attach_detach_context/attach_detach_context.dart
```diff
@@ -0,0 +1,49 @@
...
+      tracer = tp.getTracer('instrumentation-name');
+
+  final span = tracer.startSpan('root-zone-attached-span')..end();
+  final token = attach(contextWithSpan(active, span));
+
+  final completer = Completer();
+
+  // zone A
+  zoneWithContext(active).run(() {
+    final span = tracer.startSpan('zone-a-attached-span')..end();
```
dustyholmes-wf, line 22: I would advise against shadowing the variables. 

span and token are both shadowed and that certainly makes the example a little harder to read. 

One way you could tackle this is to split the zone function. The other is just using more unique names.

lib/api.dart
```diff
@@ -6,16 +6,19 @@ export 'src/api/common/resource_attributes.dart' show ResourceAttributes;
...
 export 'src/api/context/context.dart'
     show
+        active,
+        attach,
         Context,
         ContextKey,
         contextWithSpan,
         contextWithSpanContext,
+        detach,
+        root,
```
dustyholmes-wf, line 16: ❗

lib/src/api/context/context.dart
```diff
@@ -24,11 +42,68 @@ SpanContext spanContextFromContext(Context context) {
...
+  final List<ContextStackEntry> stack =
+      Zone.current[contextStackKey] ?? rootStack;
+
+  final index = stack.indexWhere((c) => c.token == token);
+  final success = index == stack.length - 1;
+  if (index != -1) {
+    stack.removeAt(index);
+  }
+
+  // TODO: log attach/detach mismatch warning
```
dustyholmes-wf, line 86: Do you intend to do this now or is this something you're saving for a future improvement.

blakeroberts-wk: tbh I hadn't really bothered thinking about it. It's easy enough to add it now

lib/api.dart
blakeroberts-wk: Do we not want this? It isn't strictly necessary, but makes it easier to create a root span vs having to construct your own context with an invalid span context

dustyholmes-wf: Nah, I'm excited to not do weird things.

lib/api.dart
```diff
@@ -6,16 +6,20 @@ export 'src/api/common/resource_attributes.dart' show ResourceAttributes;
...
     show
+        active,
+        attach,
         Context,
         ContextKey,
+        contextFromZone,
         contextWithSpan,
         contextWithSpanContext,
+        detach,
+        root,
```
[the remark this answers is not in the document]
blakeroberts-wk: Fair. I'd prefer to make the namespace optional and a prefix vs suffix. The two options I know of are:
1. Use a class with static methods
2. Use a new library file `lib/context.dart`

Would you still lean toward suffixing the function name over these other patterns? Are there other idiomatic ways to namespace functions?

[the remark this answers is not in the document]
evanweible-wf: #1 is discouraged by effective dart style guidelines (https://dart.dev/effective-dart/design#avoid-defining-a-class-that-contains-only-static-members) so I'd prefer to avoid that one. I'm open to #2, but even with a separate entrypoint, the names are still somewhat vague, especially when you consider that they are imported into the top-level namespace.

How do other language SDKs handle this?

example/attach_detach_context/attach_detach_context.dart
blakeroberts-wk: Fair. I can update the variable names. Though I'm wondering if I should scrap this example for something more contrived. Like pretend to do an app "setup" or "load" or something 🤷

lib/api.dart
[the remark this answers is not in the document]
blakeroberts-wk: Right now, I'm leaning toward making a context library and exposing context specific APIs from that library. Cross-cutting APIs like `contextWithSpan()` would remain in the API library: the context library would not depend on the API library.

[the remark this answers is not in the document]
blakeroberts-wk: Though, I hesitate. This can add another import line to some consumer files. I'm not 100% convinced of the trade off

lib/src/api/context/context.dart
```diff
@@ -1,34 +1,195 @@
...
+import 'package:logging/logging.dart';
+import 'package:meta/meta.dart';
+
 import '../../../api.dart';
 import '../trace/nonrecording_span.dart' show NonRecordingSpan;
 
+final Logger _log = Logger('opentelemetry');
+
+@sealed
 class ContextKey {}
```
blakeroberts-wk, line 15: should the constructor be made constant to allow `const myKey = ContextKey();`?

lib/api.dart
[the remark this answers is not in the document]
dustyholmes-wf: We have a Context class https://github.com/Workiva/opentelemetry-dart/pull/192/files#diff-755b890ea5c2e2edae819beea2495056b7dd283f4711c8d4a8a9c06e0640a285R176

Why not hang these off of that as statics? 

Context.root
context.active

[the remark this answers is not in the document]
dustyholmes-wf: I guess you are specifically deprecating Context.root. Why though?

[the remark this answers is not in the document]
blakeroberts-wk: DartLang discourages static members because DartLang supports top-level functions

lib/src/api/context/context_manager.dart
```diff
@@ -1,30 +1,38 @@
 // Copyright 2021-2022 Workiva.
 // Licensed under the Apache License, Version 2.0. Please see https://github.com/Workiva/opentelemetry-dart/blob/master/LICENSE for more information
 
-import '../../../api.dart';
-import '../../experimental_api.dart' show ZoneContextManager;
+import 'context.dart' show Context;
+import 'context.dart' as global show active, root;
```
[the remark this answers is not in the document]
dustyholmes-wf: Otherwise, just switch this to `import 'context.dart' show Context, active, root`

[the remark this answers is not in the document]
blakeroberts-wk: This is only temporary. This file will be removed in a future version

lib/src/api/open_telemetry.dart
```diff
@@ -41,36 +40,50 @@ void registerGlobalTextMapPropagator(api.TextMapPropagator textMapPropagator) {
 
 /// Records a span of the given [name] for the given function with a given
 /// [api.Tracer] and marks the span as errored if an exception occurs.
-@Deprecated(
-    'This method will be removed in 0.19.0. Use [traceContext] instead.')
-Future<T> trace<T>(String name, Future<T> Function() fn,
-    {api.Context? context, api.Tracer? tracer}) async {
-  context ??= api.globalContextManager.active;
-  tracer ??= _tracerProvider.getTracer('opentelemetry-dart');
+// TODO: @Deprecated('Will be removed in v0.20.0. Use [trace] instead')
```
dustyholmes-wf, line 43: I'm curious about this. 

I found with detatch that I was making use of the context provided to fn a lot.

lib/api.dart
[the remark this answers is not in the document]
dustyholmes-wf: Dartlang discourages classes that _only_ have static functions.

lib/src/api/open_telemetry.dart
blakeroberts-wk: When I had these changes before the trace helper was attaching/detaching the span it created within the zone it created, I thought it wasn't very intuitive that the argument would be the context in the zone regardless of any attached/detached context. So in the case where that context was desirable, the function argument would need to be ignored and `active` used instead. To make this an explicit conscious choice of the consumer, we could remove the argument. That way consumers would do `active` or choose the zone context specifically with `contextFromZone()`. But this is moot now that these will return the same context.

Still, the context as an argument now becomes a third API that returns the same value. If dartlang supported function parameter covariance, it'd be less of an impact to APIs. But alas, typing the callback as Function(Context) requires consumers to do `(_) {}` when favoring `active` over manual propagation.

lib/api.dart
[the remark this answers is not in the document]
blakeroberts-wk: I don't understand the benefit of that qualification

lib/src/api/open_telemetry.dart
dustyholmes-wf: I hadn't considered that I could just retrieve it from zone. Truly that didn't cross my mind.

lib/api.dart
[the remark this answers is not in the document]
dustyholmes-wf: 🤷  I'm just trying to help sort out Evan's suggestions.

lib/src/api/context/context_manager.dart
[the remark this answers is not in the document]
dustyholmes-wf: Will that version ever actually happen? I get what you are saying, but in reality I see things saying they are deprecated until the next version for years.

lib/api.dart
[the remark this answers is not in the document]
blakeroberts-wk: For sure. Like I think it is reasonable to avoid using static methods to namespace the functions since dart supports top-level functions. It's simpler and tends to be a bit more extensible. But these arguments don't change depending on whether the class is instantiated, so it feels arbitrary

lib/src/api/context/context_manager.dart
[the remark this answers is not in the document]
blakeroberts-wk: That's fair. I haven't noticed any deprecations that are blocking feature development, so I haven't seen reason to prioritize cutting v0.19.0. Still, deprecations have piled up and are hindering feature development if nothing else

lib/src/api/context/context.dart
```diff
@@ -186,10 +106,11 @@ class Context {
...
 
   /// The active context.
-  @Deprecated(
-      'This method will be removed in 0.19.0. Use the API global function '
-      '[active] instead.')
-  static Context get current => active;
+  ///
+  /// The active context is the latest attached context, if one exists, otherwise
+  /// the latest zone context, if one exists, otherwise the root context.
+  static Context get current =>
```
dustyholmes-wf, line 112: Is current more standard?

example/stream_context/stream_context.dart
```diff
@@ -8,7 +8,7 @@ import 'package:opentelemetry/sdk.dart'
     show ConsoleExporter, SimpleSpanProcessor, TracerProviderBase;
 
 mixin EventContext {
-  final Context context = active;
+  final Context context = Context.current;
```
dustyholmes-wf, line 11: I feel like this reads well where I see it througout.

blakeroberts-wk: context context context 😆

lib/src/api/context/context.dart
blakeroberts-wk: yeah. javascript is the only implementation that uses active

lib/src/api/context/context.dart
blakeroberts-wk: no. that defeats the whole purpose

lib/api.dart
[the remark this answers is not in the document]
blakeroberts-wk: Chatted async, We're going to go with statics on `Context`

bradybecker-wf (approved):

example/attach_detach_context/attach_detach_context.dart
```diff
@@ -0,0 +1,49 @@
...
+  final completer = Completer();
+
+  // zone A
+  zoneWithContext(Context.current).run(() {
+    final span = tracer.startSpan('zone-a-attached-span')..end();
+    final context = contextWithSpan(Context.current, span);
+
+    // zone B
+    zoneWithContext(context).run(() {
+      tracer.startSpan('zone-b-span').end();
```
changliu-wk, line 27: It feels like this zone-b-span should be a child of zone-a-attached-span? But the README says it is not. I suggest to use the same span names in the README.

blakeroberts-wk:
closing in favor of https://github.com/Workiva/opentelemetry-dart/pull/196

PR #194, opened 2024-10-17
O11Y-5250: Add retry

changliu-wk:

When call [client.post](https://github.com/Workiva/opentelemetry-dart/blob/master/lib/src/sdk/trace/exporters/collector_exporter.dart#L54) collector_exporter didn't check if the response is succeeded or not, this PR would like to add retry effort when the response is retryable. 

## How Has This Been Tested?

Please describe the tests that you ran to verify your change. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

## Checklist:

### DIFF
--- a/lib/src/sdk/trace/exporters/collector_exporter.dart
+++ b/lib/src/sdk/trace/exporters/collector_exporter.dart
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. Please see https://github.com/Workiva/opentelemetry-dart/blob/master/LICENSE for more information
 
 import 'dart:async';
+import 'dart:math';
 
 import 'package:fixnum/fixnum.dart';
 import 'package:http/http.dart' as http;
@@ -29,7 +30,7 @@ class CollectorExporter implements sdk.SpanExporter {
       : client = httpClient ?? http.Client();
 
   @override
-  Future<void> export(List<sdk.ReadOnlySpan> spans) async {
+  void export(List<sdk.ReadOnlySpan> spans) {
     if (_isShutdown) {
       return;
     }
@@ -38,7 +39,7 @@ class CollectorExporter implements sdk.SpanExporter {
       return;
     }
 
-    await _send(uri, spans);
+    unawaited(_send(uri, spans));
   }
 
   Future<void> _send(
@@ -47,7 +48,8 @@ class CollectorExporter implements sdk.SpanExporter {
   ) async {
     const maxRetries = 3;
     var retries = 0;
-    const valid_retry_codes = [429, 408, 500, 502, 503, 504];
+    // Retryable status from the spec: https://opentelemetry.io/docs/specs/otlp/#failures-1
+    const valid_retry_codes = [429, 502, 503, 504];
 
     final body = pb_trace_service.ExportTraceServiceRequest(
         resourceSpans: _spansToProtobuf(spans));
@@ -70,11 +72,21 @@ class CollectorExporter implements sdk.SpanExporter {
         }
       } catch (e) {
         _log.warning('Failed to export ${spans.length} spans. $e');
+        return;
       }
-      await Future.delayed(Duration(seconds: retries));
+      // Exponential backoff with jitter
+      final delay = calculateJitteredDelay(retries, Duration(seconds: 1));
+      await Future.delayed(delay);
     }
     _log.severe(
         'Failed to export ${spans.length} spans after $maxRetries retries');
+  }
+
+  Duration calculateJitteredDelay(int retries, Duration baseDelay) {
+    final random = Random();
+    final jitter = random.nextDouble() * baseDelay.inMilliseconds;
+    return Duration(
+        milliseconds: baseDelay.inMilliseconds + jitter.toInt() * retries);
   }
 
   /// Group and construct the protobuf equivalent of the given list of [api.Span]s.
--- a/lib/src/sdk/trace/exporters/collector_exporter.dart
+++ b/lib/src/sdk/trace/exporters/collector_exporter.dart
@@ -56,7 +56,7 @@ class CollectorExporter implements sdk.SpanExporter {
     final headers = {'Content-Type': 'application/x-protobuf'}
       ..addAll(this.headers);
 
-    while (retries++ < maxRetries) {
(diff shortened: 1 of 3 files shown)


aviary2-wf:
## Security Insights

## Action Items
---

lib/src/sdk/trace/exporters/collector_exporter.dart
```diff
@@ -38,23 +38,43 @@ class CollectorExporter implements sdk.SpanExporter {
       return;
     }
 
-    unawaited(_send(uri, spans));
+    await _send(uri, spans);
```
blakeroberts-wk, line 41: What is the purpose of this change?

lib/src/sdk/trace/exporters/collector_exporter.dart
```diff
@@ -38,23 +38,43 @@ class CollectorExporter implements sdk.SpanExporter {
...
-          resourceSpans: _spansToProtobuf(spans));
-      final headers = {'Content-Type': 'application/x-protobuf'}
-        ..addAll(this.headers);
-
-      await client.post(uri, body: body.writeToBuffer(), headers: headers);
-    } catch (e) {
-      _log.warning('Failed to export ${spans.length} spans.', e);
+    const maxRetries = 3;
+    var retries = 0;
+    const valid_retry_codes = [429, 408, 500, 502, 503, 504];
```
blakeroberts-wk, line 50: Here is the otel spec on retryable status codes: https://opentelemetry.io/docs/specs/otlp/#failures-1

lib/src/sdk/trace/exporters/collector_exporter.dart
changliu-wk: Letting the test case could pass.

lib/src/sdk/trace/exporters/collector_exporter.dart
```diff
@@ -38,23 +38,43 @@ class CollectorExporter implements sdk.SpanExporter {
...
+        _log.warning('Failed to export ${spans.length} spans. '
+            'HTTP status code: ${response.statusCode}');
+        // If the response is not a valid retry code, do not retry
+        if (!valid_retry_codes.contains(response.statusCode)) {
+          return;
+        }
+      } catch (e) {
+        _log.warning('Failed to export ${spans.length} spans. $e');
+      }
+      await Future.delayed(Duration(seconds: retries));
```
blakeroberts-wk, line 74: We should use exponential backoff. We should probably add jitter too, though there's implicit disagreement in the spec:
- https://opentelemetry.io/docs/specs/otel/protocol/exporter/#retry
- https://opentelemetry.io/docs/specs/otlp/#otlphttp-throttling

Both say exponential backoff, only one says jitter

lib/src/sdk/trace/exporters/collector_exporter.dart
blakeroberts-wk: I believe this is a breaking change because we'd need to update the return type of the abstract class this method overrides

changliu-wk: Reverted the change back and found a way for the test cases.

lib/src/sdk/trace/exporters/collector_exporter.dart
changliu-wk: Thanks for pointing it out.

lib/src/sdk/trace/exporters/collector_exporter.dart
changliu-wk: Added jitter in.

michaelyeager-wf (approved):

lib/src/sdk/trace/exporters/collector_exporter.dart
```diff
@@ -45,16 +46,47 @@ class CollectorExporter implements sdk.SpanExporter {
...
+      // Exponential backoff with jitter
+      final delay = calculateJitteredDelay(retries, Duration(seconds: 1));
+      await Future.delayed(delay);
     }
+    _log.severe(
+        'Failed to export ${spans.length} spans after $maxRetries retries');
+  }
+
+  Duration calculateJitteredDelay(int retries, Duration baseDelay) {
+    final random = Random();
```
blakeroberts-wk, line 86: Can we make this a class member variable?

lib/src/sdk/trace/exporters/collector_exporter.dart
```diff
@@ -45,16 +46,47 @@ class CollectorExporter implements sdk.SpanExporter {
...
+    final jitter = random.nextDouble() * baseDelay.inMilliseconds;
+    return Duration(
+        milliseconds: baseDelay.inMilliseconds + jitter.toInt() * retries);
```
blakeroberts-wk, line 89: We should use exponential delay, something like this:
```suggestion
    final delay = baseDelay.inMilliseconds * pow(2, retries)
    final jitter = random.nextDouble() * delay;
    return Duration(milliseconds: (delay + jitter).toInt());
```

lib/src/sdk/trace/exporters/collector_exporter.dart
```diff
@@ -45,16 +46,47 @@ class CollectorExporter implements sdk.SpanExporter {
...
+        // If the response is not a valid retry code, do not retry
+        if (!valid_retry_codes.contains(response.statusCode)) {
+          return;
+        }
+      } catch (e) {
+        _log.warning('Failed to export ${spans.length} spans. $e');
+        return;
+      }
+      // Exponential backoff with jitter
+      final delay = calculateJitteredDelay(retries, Duration(seconds: 1));
```
blakeroberts-wk, line 78: I think it's reasonable to start with something smaller like 100 milliseconds

lib/src/sdk/trace/exporters/collector_exporter.dart
blakeroberts-wk: This assumes the first retry will have "retries" set to zero, not one

blakeroberts-wk (approved):

changliu-wk:
@Workiva/release-management-p

changliu-wk:
QA +1

rmconsole-wf (approved):
+1 from RM

PR #197, opened 2024-11-26
O11Y-5307: Run CI Once

kennytrytek-wf:


.github/workflows/dart_ci.yaml
```diff
@@ -1,13 +1,15 @@
...
 on:
+  workflow_dispatch:
   push:
     branches:
-      - 'master'
-      - 'test_consume_*'
+      - "master"
+      - "_integration/**/*"
+      - "experiments/**/*"
+      - "test_consume_**"
```
blakeroberts-wk, line 9: Are any of these used at all or should we just do master?

kennytrytek-wf: No, they are not used, but it doesn't hurt anything to keep them available for use if somebody wants to use them. We already set it up this way for the other 34 repos.

kennytrytek-wf: That's fine, opentelemetry-dart has its own use-cases that differ from other O11y repos.

blakeroberts-wk (approved):

kennytrytek-wf:
QA +1 ⛄ 

@Workiva/release-management-pp

rmconsole-wf (approved):
+1 from RM

PR #200, opened 2025-01-07
add sampling support to span processors

blakeroberts-wk:

This adds sampling support to the simple and batch span processors. As it stands now, even if a sampler is configured, the sampling decision does not affect whether the span is exported.

Fixes # N/A

There are two functional changes:
1. updating the batch and simple processors to check the sampling flag
2. updating the trace helper functions to support taking a list of attributes to add at span start necessary for sampling on attributes

Additionally:
1. an example program was added
2. unit tests were updated
3. exporter's forceFlush was deprecated because it's a kind of state dependent operation that should be managed by the processor that uses an exporter.
4. non-functional changes to simplify the `SpanContext` class.

## How Has This Been Tested?

The example program and unit tests sufficiently demonstrate functionality.

## Checklist:


blakeroberts-wk:
QA +1 the example program outputs the following:

```
Span started: span-not-sampled
Span ended: span-not-sampled
Span started: span-sampled
Span ended: span-sampled
{traceId: 94f26b7841f0aa268128fd1f4e4f493c, parentId: 0000000000000000, name: span-sampled, id: ac98e8435956ac7e, timestamp: 1736286063230117000, duration: 2624000, flags: 01, state: , status: StatusCode.unset}
Shutting down
```

todbachman-wf:
**Question**: I'm concerned about the long term ownership and maintainability of this code. This feels like a lot of very specific observability code to add to wdesk_sdk given our team doesn't own the observability story and doesn't/won't understand this code well enough to own it and maintain it long term. This has historically introduced challenges for our team as others have put code in wdesk_sdk because it's a convenient place to put it. Can we instead house this code in a repo that the Observability team owns and expose a public API that wdesk_sdk can then consume?

evanweible-wf (approved):

blakeroberts-wk:
QA +1

blakeroberts-wk:
@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #202, opened 2025-01-09
feat: proto logs

yuzurihaaa:
## Which problem is this PR solving?

To add proto generated dart files. Next PR will be to create API for log.

## Short description of the change
Add generated proto files.

## How Has This Been Tested?

Please describe the tests that you ran to verify your change. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration
1. Ran unit tests

## Checklist:

- [ ] Unit tests have been added
- [ ] Documentation has been updated

PR #203, opened 2025-01-10
feat: logs api

yuzurihaaa:
1. This is a continuation of https://github.com/Workiva/opentelemetry-dart/pull/202. Please Review that PR first before continuing on this PR due to huge change on proto generated code from the mentioned PR.

Fixes # (issue)

1. Add OpenTelemetry Logs API layer.
2. I am trying to make it as close as possible to https://www.npmjs.com/package/@opentelemetry/api-logs 
3. I export them in `experimental_api.dart` as in js, this feature is still in experimental, but `experimental_api.dart` is placed under `lib/src/` instead of `lib/` which will require defining import `package:opentelemetry/lib/src/experimental_api.dart` when need it. Let me know if I should place them under `sdk.dart`

## How Has This Been Tested?

1. Unit test.

## Checklist:


yuzurihaaa:
branch rebased from latest `master`

lib/src/api/logs/log_record.dart
```diff
@@ -0,0 +1,102 @@
+// Copyright 2021-2022 Workiva.
+// Licensed under the Apache License, Version 2.0. Please see https://github.com/Workiva/opentelemetry-dart/blob/master/LICENSE for more information
+
+import 'package:opentelemetry/api.dart' as api;
+import 'package:opentelemetry/sdk.dart' as sdk;
```
blakeroberts-wk, line 5: nit: use relative imports

lib/src/api/logs/logger_provider.dart
```diff
@@ -0,0 +1,25 @@
...
+  /// The [version] specifies the version of the instrumentation scope if the
+  /// scope has a version (e.g. a library version). The [schemaUrl] identifies
+  /// the schema this provider adheres to.  The [attributes] specifies
+  /// attributes to associate with emitted telemetry.
+  Logger get(
+    String name, {
+    String version = '',
+    String schemaUrl = '',
+    List<Attribute> attributes = const [],
+    bool? includeTraceContext,
```
blakeroberts-wk, line 23: What is the intended purpose of `includeTraceContext`? I can't find this in the specification for a logger: https://opentelemetry.io/docs/specs/otel/logs/api/#logger

lib/src/api/logs/logger.dart
```diff
@@ -0,0 +1,8 @@
+// Copyright 2021-2022 Workiva.
+// Licensed under the Apache License, Version 2.0. Please see https://github.com/Workiva/opentelemetry-dart/blob/master/LICENSE for more information
+
+import 'log_record.dart';
+
+abstract class Logger {
+  void emit(LogRecord logRecord);
```
blakeroberts-wk, line 7: I believe we want this method to take the individual parameters listed [here](https://opentelemetry.io/docs/specs/otel/logs/api/#emit-a-logrecord) and the implementation creates a log record from the parameters.

lib/src/api/common/export_result.dart
```diff
@@ -0,0 +1,15 @@
+// Copyright 2021-2022 Workiva.
+// Licensed under the Apache License, Version 2.0. Please see https://github.com/Workiva/opentelemetry-dart/blob/master/LICENSE for more information
+
+class ExportResult {
```
blakeroberts-wk, line 4: If this is intended to be used by an exporter, I believe we'll want this defined within the SDK.

yuzurihaaa: Reason why I put under `common` because it is placed in under `core` in [opentelemetry-js](https://github.com/open-telemetry/opentelemetry-js/blob/fc0edd82f6749bfe1ca733ed1f80ad4a6015285f/packages/opentelemetry-core/src/ExportResult.ts#L22) 

But I agree with u, let me remove this and put in `SDK` branch.

lib/src/api/logs/log_record.dart
yuzurihaaa: Remove after changes related to https://github.com/Workiva/opentelemetry-dart/pull/203#discussion_r1915329075

lib/src/api/logs/logger.dart
yuzurihaaa: Updated. I initially wanted to make the object to initialise `Context` if no value is given, but I think I found a way to achieve that. Your suggestion makes this api better :)

yuzurihaaa: Updated

lib/src/api/logs/logger_provider.dart
yuzurihaaa: Updated

lib/src/api/common/export_result.dart
yuzurihaaa: Updated

yuzurihaaa:
Hi @blakeroberts-wk , about unit test, seems like the part that is causing it to fail is 

```
warning • lib/src/api/context/context.dart:22:18 • The declaration '_contextStackKey' isn't referenced. Try removing the declaration of '_contextStackKey'. •
          unused_element
```

I ran `make analyze` on local and I got the same error `make: *** [analyze] Error 2`

removing `final ContextKey _contextStackKey = ContextKey();` , run `make analyze` and the error no more. Should I add the change here to make the test pass or another PR?

yuzurihaaa:
rebased with latest master

blakeroberts-wk (approved):

blakeroberts-wk:
QA +1 experimental public or private API additions only, CI is sufficient

rmconsole-wf (approved):
+1 from RM

PR #204, opened 2025-01-10
feat: log sdk

yuzurihaaa:
## Which problem is this PR solving?
1. This is a continuation of https://github.com/Workiva/opentelemetry-dart/pull/202 and https://github.com/Workiva/opentelemetry-dart/pull/203. Please Review them first before continuing on this PR.

Fixes # (issue)

## Short description of the change
1. Add OpenTelemetry Logs SDK.
2. I am trying to make it as close as possible to https://www.npmjs.com/package/@opentelemetry/api-logs.

## How Has This Been Tested?

1. Unit test.

## Checklist:

- [x] Unit tests have been added
- [ ] Documentation has been updated

PR #207, opened 2025-01-16
feat: opentelemetry sdk LoggerProvider

yuzurihaaa:
1. To add LoggerProvider in SDK.
2. This is from [Log SDK PR](https://github.com/Workiva/opentelemetry-dart/pull/204) but broken down to be smaller. This PR only for LoggerProvider and related parameters to ease review.

## How Has This Been Tested?
1. Unit test

## Checklist:


lib/src/sdk/common/limits.dart
```diff
@@ -77,6 +78,24 @@ api.Attribute applyAttributeLimits(api.Attribute attr, sdk.SpanLimits limits) {
   return attr;
 }
 
+@protected
+api.Attribute applyAttributeLimitsForLog(api.Attribute attr, LogRecordLimits limits) {
+  // if maxNumAttributeLength is less than zero, then it has unlimited length.
+  if (limits.attributeValueLengthLimit < 0) return attr;
+
+  if (attr.value is String) {
+    attr = api.Attribute.fromString(
```
blakeroberts-wk, line 87: We should avoid allocating a new attribute if limits weren't exceeded.

lib/src/sdk/logs/log_record.dart
```diff
@@ -0,0 +1,189 @@
...
+class LogRecord implements ReadWriteLogRecord {
+  @override
+  final sdk.InstrumentationScope instrumentationScope;
+
+  @override
+  final sdk.Resource? resource;
+
+  final sdk.TimeProvider _timeProvider;
+  final api.Context? context;
+  final api.SpanContext? _spanContext;
```
blakeroberts-wk, line 54: Do we need references to both of these? Can we make context a non-null private member initialized to current if the optional parameter isn't given? And then remove _spanContext and use _context to retrieve a spanContext?

lib/src/sdk/logs/log_record.dart
```diff
@@ -0,0 +1,189 @@
...
+
+  bool _isReadonly = false;
+  String? _severityText;
+  api.Severity? _severityNumber;
+  dynamic _body;
+  int _totalAttributesCount = 0;
+
+  final sdk.Attributes _attributes;
+
+  LogRecord({
```
blakeroberts-wk, line 67: Should this be a private constructor?

lib/src/sdk/logs/log_record.dart
```diff
@@ -0,0 +1,189 @@
...
+
+  @override
+  int get droppedAttributesCount => _totalAttributesCount - (attributes?.length ?? 0);
+
+  @override
+  Int64? get timeStamp => _timeStamp != null ? Int64(_timeStamp!.microsecondsSinceEpoch) * 1000 : _timeProvider.now;
+
+  @override
+  Int64? get observedTimestamp =>
+      _observedTimestamp != null ? Int64(_observedTimestamp!.microsecondsSinceEpoch) * 1000 : _timeProvider.now;
```
blakeroberts-wk, line 113: Does it make sense to avoid using the int64 type until exporting OTLP format?

lib/src/sdk/logs/log_record.dart
```diff
@@ -0,0 +1,189 @@
...
+    _severityText = severity;
+  }
+
+  void setAttributes(sdk.Attributes attributes) {
+    for (final key in attributes.keys) {
+      setAttribute(key, attributes.get(key));
+    }
+  }
+
+  void setAttribute(String key, Object? value) {
```
blakeroberts-wk, line 139: Should this take an attribute class to allow consumers to save memory by reusing attributes?

lib/src/sdk/logs/log_record.dart
```diff
@@ -0,0 +1,189 @@
...
+  @override
+  String? get severityText => _severityText;
+
+  @override
+  set severityText(String? severity) {
+    if (_isReadonly) return;
+    _severityText = severity;
+  }
+
+  void setAttributes(sdk.Attributes attributes) {
```
blakeroberts-wk, line 133: Should we take a list of attributes for easier integration?

lib/src/sdk/logs/logger.dart
```diff
@@ -0,0 +1,48 @@
...
+import 'package:opentelemetry/src/experimental_sdk.dart' as sdk;
+
+class Logger extends api.Logger {
+  final sdk.InstrumentationScope instrumentationScope;
+  final sdk.Resource? resource;
+  final Function(sdk.ReadWriteLogRecord)? onLogEmit;
+  final sdk.LogRecordLimits logRecordLimits;
+  final sdk.TimeProvider? timeProvider;
+
+  Logger({
```
blakeroberts-wk, line 16: Should this be a private constructor?

lib/src/sdk/logs/logger.dart
```diff
@@ -0,0 +1,48 @@
...
+      resource: resource,
+      instrumentationScope: instrumentationScope,
+      context: context,
+      severityText: severityText,
+      severityNumber: severityNumber,
+      attributes: attributes,
+      body: body,
+      timeProvider: timeProvider,
+    );
+    onLogEmit?.call(log);
```
blakeroberts-wk, line 45: Should we be getting a list of processors from the provider?

lib/src/sdk/logs/logger.dart
```diff
@@ -0,0 +1,48 @@
...
+    log.makeReadonly();
```
blakeroberts-wk, line 46: Would it be cleaner to have two types where one is immutable and the other is mutable?

lib/src/sdk/logs/logger_provider.dart
```diff
@@ -0,0 +1,78 @@
...
+import 'package:meta/meta.dart';
+import 'package:opentelemetry/api.dart' as api;
+import 'package:opentelemetry/sdk.dart' as sdk;
+import 'package:opentelemetry/src/experimental_api.dart' as api;
+import 'package:opentelemetry/src/experimental_sdk.dart' as sdk;
+import 'package:opentelemetry/src/sdk/logs/log_record_limit.dart';
+import 'package:opentelemetry/src/sdk/logs/logger_config.dart';
+import 'package:quiver/core.dart';
+
+const defaultLoggerName = 'unknown';
```
blakeroberts-wk, line 13: Should this be a default scope with a name set to opentelemetry?

lib/src/sdk/logs/logger_provider.dart
```diff
@@ -0,0 +1,78 @@
...
+  })  : processors = processors ?? <sdk.LogRecordProcessor>[],
+        _timeProvider = timeProvider ?? sdk.DateTimeTimeProvider();
+
+  @override
+  api.Logger get(
+    String name, {
+    String version = '',
+    String schemaUrl = '',
+    List<api.Attribute> attributes = const [],
+    bool? includeTraceContext,
```
blakeroberts-wk, line 44: Is this needed?

lib/src/sdk/common/limits.dart
yuzurihaaa: Hi @blakeroberts-wk , let me check if I understand you correctly. Do you mean that if the attribute value between that is passed in `Attribute` and `truncated` is the same, return the `Attribute` but if they are the different, return the newly created `Attribute`?

instead of 
```
attr = api.Attribute.fromString(
        attr.key, applyAttributeLengthLimit(attr.value as String, limits.attributeValueLengthLimit));
```
do
```
 final truncatedValue = applyAttributeLengthLimit(
        attr.value as String, limits.attributeValueLengthLimit);

    if (truncatedValue == attr.value) return attr;

    return api.Attribute.fromString(attr.key, truncatedValue);
```

lib/src/sdk/logs/log_record.dart
yuzurihaaa: I will still need it in `logger.dart`.
```
@override
  void emit({
    sdk.Attributes? attributes,
    Context? context,
    dynamic body,
    DateTime? observedTimestamp,
    api.Severity? severityNumber,
    String? severityText,
    DateTime? timeStamp,
  }) {
    final log = sdk.LogRecord(
      logRecordLimits: logRecordLimits,
      resource: resource,
      instrumentationScope: instrumentationScope,
      context: context,
      severityText: severityText,
      severityNumber: severityNumber,
      attributes: attributes,
      body: body,
      timeProvider: timeProvider,
    );
    onLogEmit?.call(log);
    log.makeReadonly();
  }
```

If this is changed to private constructor, I won't be able to use it there.

lib/src/sdk/logs/log_record.dart
yuzurihaaa: Are you suggesting to use `int` (Dart primitive type) and convert to `Int64` later on or use `DateTime` here?

lib/src/sdk/logs/log_record.dart
yuzurihaaa: not too sure about save memory because `Attributes` already in `final Map<String, Object> _attributes = {};`

and calling `add` function will call

```
void add(Attribute attribute) {
  _attributes[attribute.key] = attribute.value;
}
```.

Isn't the memory consumption will always stay the same as long as the attribute key is same?

File: lib/src/sdk/logs/logger.dart (hunk 7 above)
Comment (yuzurihaaa) [reply lib/src/sdk/logs/logger.dart:16]: I don't think so. This class is called in `LoggerProvider`

```
loggers.putIfAbsent(
      key,
      () => sdk.Logger(
        logRecordLimits: logRecordLimits,
        resource: resource,
        instrumentationScope: sdk.InstrumentationScope(loggerName, version, schemaUrl, attributes),
        timeProvider: _timeProvider,
        onLogEmit: (log) {
          for (final processor in processors) {
            processor.onEmit(log);
          }
        },
      ),
    );
```

File: lib/src/sdk/logs/logger_provider.dart (hunk 11 above)
Comment (yuzurihaaa) [reply lib/src/sdk/logs/logger_provider.dart]: removed

File: lib/src/sdk/logs/logger_provider.dart (hunk 10 above)
Comment (yuzurihaaa) [reply lib/src/sdk/logs/logger_provider.dart:13]: As I don't have any reference for this default value in Tracer, so I use https://github.com/open-telemetry/opentelemetry-js/blob/main/experimental/packages/sdk-logs/src/LoggerProvider.ts#L29 as reference.

Comment (yuzurihaaa) [reply lib/src/sdk/logs/logger_provider.dart]: Changed to `opentelemetry`

File: lib/src/sdk/logs/logger_provider.dart [hunk 12]
```diff
@@ -0,0 +1,78 @@
...
+            processor.onEmit(log);
+          }
+        },
+      ),
+    );
+  }
+
+  void addLogRecordProcessor(sdk.LogRecordProcessor processor) {
+    processors.add(processor);
+  }
```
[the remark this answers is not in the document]
yuzurihaaa: yes can, but then next question will be, should `processors` be exported by `LoggerProvider`? For now I change to be immutable first.

lib/src/sdk/logs/processors/noop_log_processor.dart
```diff
@@ -0,0 +1,17 @@
+// Copyright 2021-2022 Workiva.
+// Licensed under the Apache License, Version 2.0. Please see https://github.com/Workiva/opentelemetry-dart/blob/master/LICENSE for more information
+
+import 'package:opentelemetry/src/experimental_sdk.dart' as sdk;
+
+class NoopLogRecordProcessor implements sdk.LogRecordProcessor {
```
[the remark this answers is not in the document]
yuzurihaaa: Not really, but from opentelemetry-js it is there https://github.com/open-telemetry/opentelemetry-js/blob/15ba2d7da3bd6663ee6008e189f004bab93ff6fe/experimental/packages/sdk-logs/src/export/NoopLogRecordProcessor.ts#L21 so I put it for parity.

Let me know if I should remove it.

lib/src/sdk/logs/processors/log_record_processor.dart
```diff
@@ -0,0 +1,13 @@
+// Copyright 2021-2022 Workiva.
+// Licensed under the Apache License, Version 2.0. Please see https://github.com/Workiva/opentelemetry-dart/blob/master/LICENSE for more information
+
+import 'package:opentelemetry/src/experimental_sdk.dart' as sdk;
+
+/// https://opentelemetry.io/docs/specs/otel/logs/sdk/#logrecordprocessor
+abstract class LogRecordProcessor {
+  void onEmit(sdk.ReadableLogRecord logRecord);
```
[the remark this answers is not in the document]
yuzurihaaa: `ReadableLogRecord` is the base abstract class. The processor can infer the type to be `ReadWriteLogRecord`. As long as the class is implemented as `ReadableLogRecord` then it is a valid `LogRecord`.

The sub class to this is the `ReadWriteLogRecord` which has setter for some parameters. For customisation, as long as the `LogRecord` adhere to `ReadableLogRecord` then it is good to use.

lib/src/sdk/logs/logger_provider.dart
```diff
@@ -0,0 +1,78 @@
...
+
+  Future<void> forceFlush() async {
+    await Future.forEach(processors, (e) => e.forceFlush());
+  }
+
+  Future<void> shutdown() async {
+    await Future.forEach(processors, (e) => e.shutdown());
+  }
```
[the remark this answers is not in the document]
yuzurihaaa: `LogRecordProcessor` is implemented with

```
Future<void> forceFlush();

Future<void> shutdown();
```

it would only make sense to use `async` here to ensure all the processors are done with their respective processes.

Also, with `async` it provides the ability for graceful shutdown for time-consuming operation.

lib/src/sdk/logs/logger_provider.dart
```diff
@@ -0,0 +1,78 @@
...
+
+  final List<sdk.LogRecordProcessor> processors;
+
+  final sdk.Resource? resource;
+  final sdk.LogRecordLimits logRecordLimits;
+
+  final sdk.TimeProvider _timeProvider;
+
+  LoggerProvider({
+    this.resource,
```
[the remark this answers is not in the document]
yuzurihaaa: Change to optional with assigning default value, as `TracerProviderBase` did it.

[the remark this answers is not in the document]
yuzurihaaa: I also change all the variables to be private here as I am not sure if needed from anywhere else

lib/src/sdk/logs/logger.dart
yuzurihaaa: U mean to pass the processors here? For me they serve the same purpose, so I don't really have a preference.

Referring from https://github.com/open-telemetry/opentelemetry-java/blob/19e964a6368452e59a5c6b3488eaf5b0f60ef2e4/sdk/logs/src/main/java/io/opentelemetry/sdk/logs/SdkLogRecordBuilder.java#L128-L143

https://github.com/open-telemetry/opentelemetry-js/blob/15ba2d7da3bd6663ee6008e189f004bab93ff6fe/experimental/packages/sdk-logs/src/Logger.ts#L49

actually both java and js is using something like `SharedState` class, but here I skip that and pass the values instead. Let me know if I should change for parity.

lib/src/sdk/logs/logger.dart
yuzurihaaa: The intention of this `makeReadonly` to allow mutation (if any) only during `onLogEmit?.call`.

lib/src/sdk/logs/log_record.dart
yuzurihaaa: Updated

lib/src/sdk/logs/log_record.dart
yuzurihaaa: Sure. Updated.

lib/src/sdk/common/limits.dart
yuzurihaaa: pushed the change for this. Let me know if I got this wrong.

lib/src/sdk/logs/log_record.dart
yuzurihaaa: Updated to `api.Attribute` after changes for https://github.com/Workiva/opentelemetry-dart/pull/207#discussion_r1935800595

lib/src/sdk/logs/log_record.dart
blakeroberts-wk: I think DateTime makes sense.

lib/src/sdk/logs/log_record.dart
blakeroberts-wk: It's more about the allocation than the retained size. But either way, consumers can create a final attribute and reuse it on multiple different spans

lib/src/sdk/logs/log_record.dart
yuzurihaaa: I need some feedback for using `DateTime`. Currently I am injecting `DateTimeTimeProvider`, but with DateTime, should I change `LogRecord` constructor to be 
```dart
LogRecord({
...,
    DateTime? timeStamp,
    DateTime? observedTimestamp,
}): 
_timeStamp = timeStamp ?? DateTime.now(),
_observedTimestamp = observedTimestamp ?? DateTime.now(),
```

or keep it `null` and use `DateTime? get timeStamp => _timeStamp ?? DateTime.fromMicrosecondsSinceEpoch((_timeProvider.now ~/ 1000).toInt());` in the getter?

I also would like you to verify if the logic here is correct.

lib/src/sdk/logs/log_record.dart
yuzurihaaa: I've already updated this in recent change, can I know if the change is the expected or I missed something out here?

yuzurihaaa:
Hi @blakeroberts-wk , I saw that you've resolve some of the comments, but there are still some that are not resolved but I didn't see any feedback / response.

lib/src/sdk/logs/logger_provider.dart
```diff
@@ -0,0 +1,80 @@
...
+    sdk.Resource? resource,
+    List<sdk.LogRecordProcessor>? processors,
+    sdk.TimeProvider? timeProvider,
+  })  : _processors = processors ?? const <sdk.LogRecordProcessor>[],
+        _config = config,
+        _logRecordLimits = logRecordLimits,
+        _resource = resource ?? sdk.Resource([]),
+        _timeProvider = timeProvider ?? sdk.DateTimeTimeProvider();
+
+  UnmodifiableListView<sdk.LogRecordProcessor> get processors =>
```
blakeroberts-wk, line 41: Rather than having a public getter that returns an unmodifiable list, could we reduce the scope and remove the wrapper?

lib/src/sdk/logs/logger_provider.dart
```diff
@@ -0,0 +1,80 @@
...
+    List<api.Attribute> attributes = const [],
+  }) {
+    final loggerName = name.isNotEmpty ? name : defaultLoggerName;
+    final key = hash3(loggerName, version, schemaUrl);
+    if (_config.disabled) {
+      return api.NoopLogger();
+    }
+    return _loggers.putIfAbsent(
+      key,
+      () => sdk.Logger(
```
blakeroberts-wk, line 58: If we switch to a factory with package private scope, can this constructor be made private?

lib/src/sdk/logs/logger_provider.dart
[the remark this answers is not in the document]
blakeroberts-wk: I'm unsure if we want to return a future here. For example, the browser's sendBeacon() doesn't return a promise. Even so, I am sure we don't need the async or await keyword. Those can be safely removed

lib/src/sdk/logs/logger_provider.dart
yuzurihaaa: Removed

yuzurihaaa:
https://github.com/Workiva/opentelemetry-dart/pull/207#discussion_r1951172171

not sure why I can't reply here. Removed `async` `await` keyword. 

> For example, the browser's sendBeacon() doesn't return a promise

This is actually something that I am unaware of and not really sure how to handle for graceful shutdown if there are async process going on, but the function only return `void`. I've changed to `FutureOr` for better readability.

lib/src/sdk/logs/logger_provider.dart
yuzurihaaa: > If we switch to a factory with package private scope

I am sorry but I don't understand what do you mean here.

Currently the constructor is
```
Logger({
    required this.instrumentationScope,
    required this.logRecordLimits,
    this.onLogEmit,
    this.resource,
    this.timeProvider,
  });
```

but if I changed it to `Logger._(...` and `factory Logger.create(...` then I will still pass the same arguments so I don't see the benefit of it. Maybe I don't fully understand the statement here.

lib/src/sdk/logs/logger_provider.dart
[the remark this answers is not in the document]
yuzurihaaa: Removed `async` `await` keyword. 

> For example, the browser's sendBeacon() doesn't return a promise

This is actually something that I am unaware of and not really sure how to handle for graceful shutdown if there are async process going on, but the function only return `void` as the processes are still on oing. I've changed to `FutureOr` for better readability.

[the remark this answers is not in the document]
blakeroberts-wk: > The navigator.sendBeacon() method [asynchronously](https://developer.mozilla.org/en-US/docs/Glossary/Asynchronous) sends an [HTTP POST](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/POST) request containing a small amount of data to a web server.
>
> It's intended to be used for sending analytics data to a web server, and avoids some of the problems with legacy techniques for sending analytics, such as the use of [XMLHttpRequest](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest).
>
> Note: For use cases that need the ability to send requests with methods other than POST, or to change any request properties, or that need access to the server response, instead use the [fetch()](https://developer.mozilla.org/en-US/docs/Web/API/Window/fetch) method with [keepalive](https://developer.mozilla.org/en-US/docs/Web/API/RequestInit#keepalive) set to true.

https://developer.mozilla.org/en-US/docs/Web/API/Navigator/sendBeacon

I think it makes sense to mimic the send beacon API such that the API is synchronous but internally manages some async processes. For example, an exporter will internally need to watch responses and potentially retry requests. But it can return synchronously as soon as the request is queued (with "keepalive").

And if callers of the synchronous API must execute the code asynchronously, they can wrap the synchronous function however they'd like (micro task or event loop) such as `Future(() => mySyncFunc())`.

To speak to graceful shutdown specifically: most likely, shutdown will be called via a listener registered to an app shutdown event. In this case, returning a future doesn't make sense. This would imply that the caller (the event listener) needs to await the future to accomplish graceful shutdown. And this await would be similarly superfluous to the async/await you removed from within these functions: we just make the function synchronous resulting in the removal of the async/await that would otherwise be necessary for the listener. The synchronous shutdown should be implemented similar to sendBeacon: force flush pending data using xhr with keepalive and without waiting for the response. But this is an implementation detail that can be solely contained within an exporter's implementation.

lib/src/sdk/logs/log_record.dart
blakeroberts-wk: I was answering your question. We can resolve this thread.

lib/src/sdk/logs/log_record.dart
blakeroberts-wk: What we want is:
- use timestamps from `performance.now()` when developing for the web
- use timestamps from `Stopwatch` (or similar) when developing for the VM
- do not accept or return `Int64` types leaving this as an internal implementation detail of an otlp exporter

To support both platforms, the shared type we want to use might be `double`. And we should align this type with what is returned by a time provider. Of course changing the return type of the time provider interface method is a breaking change. So we'd need to mark the method as deprecated stating that the return type will change in the next minor version.

lib/src/sdk/logs/logger_provider.dart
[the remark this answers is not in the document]
yuzurihaaa: Got it. In that case, I'll change it to `void` and will keep this in mind for shutdown process that requires for await

lib/src/sdk/logs/logger_provider.dart
yuzurihaaa: removed

lib/src/sdk/logs/logger_provider.dart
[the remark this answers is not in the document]
yuzurihaaa: resolving this as the getter is not removed and processors is set to private.

yuzurihaaa:
Hi @blakeroberts-wk , been a while. Can I know if there are any issues that can be resolved. There are also issues that I need some help :)

lib/src/sdk/logs/logger.dart
blakeroberts-wk: In this case, I'm not sure it makes sense to follow either of those implementations. I imagine the designs above are due to legacy reasons or oversight. We should support multiple processors. Those other implementations accomplish this by introducing a `MultiLogRecordProcessor`. As far as I can tell, this is superfluous complexity. A logger provider should have a list of processors given to a logger during instantiation. This implementation follows the same lines as other providers (meter and tracer).

https://opentelemetry.io/docs/specs/otel/logs/sdk/#logrecordprocessor

lib/src/sdk/logs/logger.dart
```diff
@@ -0,0 +1,49 @@
...
+
+import 'package:opentelemetry/sdk.dart' as sdk;
+import 'package:opentelemetry/api.dart' as api;
+import 'package:opentelemetry/src/api/context/context.dart';
+import 'package:opentelemetry/src/experimental_api.dart' as api;
+import 'package:opentelemetry/src/experimental_sdk.dart' as sdk;
+
+class Logger extends api.Logger {
+  final sdk.InstrumentationScope instrumentationScope;
+  final sdk.Resource? resource;
```
blakeroberts-wk, line 12:
```suggestion
  final sdk.Resource resource;
```

lib/src/sdk/logs/logger.dart
```diff
@@ -0,0 +1,49 @@
...
+  final Function(sdk.ReadWriteLogRecord)? onLogEmit;
```
blakeroberts-wk, line 13:
```suggestion
  final Function(sdk.ReadWriteLogRecord) onLogEmit;
```

lib/src/sdk/logs/logger.dart
```diff
@@ -0,0 +1,49 @@
...
+  final sdk.LogRecordLimits logRecordLimits;
+  final sdk.TimeProvider? timeProvider;
```
blakeroberts-wk, line 15:
```suggestion
  final sdk.TimeProvider timeProvider;
```

yuzurihaaa: updated

lib/src/sdk/logs/logger.dart
yuzurihaaa: updated

lib/src/sdk/logs/logger.dart
yuzurihaaa: updated

lib/src/sdk/logs/log_record.dart
yuzurihaaa: Is the annotate the constructor with `@protected` is the goal for `private constructor` or this is a simplified version? Actually I am really looking forward to know the answer for the `private constructor` here, since I can't think of any.

yuzurihaaa: Is the annotate the constructor with `@protected` is the goal for `private constructor` or this is a simplified version? Actually I am really looking forward to know the answer for the `private constructor` here, since I can't think of any.

lib/src/sdk/logs/processors/log_record_processor.dart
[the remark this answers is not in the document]
blakeroberts-wk: > The processor can infer the type to be ReadWriteLogRecord.

Type inferencing does not apply here. The processor implementation can type check/cast, but this isn't inferring the type. And since a processor is allowed to mutate the record on emit, we should write the contract this way (define the API to use a mutable record). Otherwise, implementations must type check/cast to perform mutations

lib/src/sdk/logs/logger_provider.dart
blakeroberts-wk: Annotating the constructor with `@protected` should be sufficient

lib/src/sdk/common/limits.dart
yuzurihaaa: Updated

lib/src/sdk/logs/log_record.dart
yuzurihaaa: updated

lib/src/sdk/logs/logger.dart
yuzurihaaa: Updated

lib/src/sdk/logs/processors/noop_log_processor.dart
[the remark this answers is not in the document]
yuzurihaaa: Updated

lib/src/sdk/logs/logger.dart
yuzurihaaa: Got it, changed from callback to list of processors here

lib/src/sdk/logs/log_record_limit.dart
```diff
@@ -0,0 +1,28 @@
+// Copyright 2021-2022 Workiva.
+// Licensed under the Apache License, Version 2.0. Please see https://github.com/Workiva/opentelemetry-dart/blob/master/LICENSE for more information
+
+// https://opentelemetry.io/docs/specs/otel/logs/sdk/#logrecord-limits
+abstract class LogRecordLimits {
```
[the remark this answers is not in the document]
yuzurihaaa: Removed, changed to `class`

lib/src/sdk/logs/log_record.dart
```diff
@@ -0,0 +1,147 @@
...
+
+import 'package:meta/meta.dart';
+import 'package:opentelemetry/api.dart' as api;
+import 'package:opentelemetry/sdk.dart' as sdk;
+import 'package:opentelemetry/src/experimental_api.dart' as api;
+import 'package:opentelemetry/src/experimental_sdk.dart' as sdk;
+import 'package:opentelemetry/src/sdk/common/limits.dart';
+
+/// https://opentelemetry.io/docs/specs/otel/logs/sdk/#readwritelogrecord
+abstract class ReadableLogRecord {
```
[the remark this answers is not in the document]
yuzurihaaa: Updated all values to be non-nullable

lib/src/sdk/common/limits.dart
```diff
@@ -77,6 +79,36 @@ api.Attribute applyAttributeLimits(api.Attribute attr, sdk.SpanLimits limits) {
...
+    final listString = attr.value as List<String>;
+    final truncatedValues = listString
+        .map((e) =>
+            applyAttributeLengthLimit(e, limits.attributeValueLengthLimit))
+        .toList();
+
+    final equal = const ListEquality().equals(listString, truncatedValues);
+    if (equal) return attr;
+
+    return api.Attribute.fromStringList(attr.key, truncatedValues);
```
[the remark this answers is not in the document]
yuzurihaaa: > This does not avoid allocations because of map().toList()

Learning something new :)

Updated

lib/src/sdk/logs/log_record.dart
yuzurihaaa: Let me try to sort this one out too then

can I know if this is the expected change?

```diff
abstract class TimeProvider {
  // The smallest increment that DateTime can report is in microseconds, while
  // OpenTelemetry expects time in nanoseconds.
  @Deprecated('This constant will be removed in 0.19.0 without replacement.')
  static const int nanosecondsPerMicrosecond = 1000;

  // window.performance API reports time in fractional milliseconds, while
  // OpenTelemetry expects time in nanoseconds.
  @Deprecated('This constant will be removed in 0.19.0 without replacement.')
  static const int nanosecondsPerMillisecond = 1000000;

  /// The current time, in nanoseconds since Unix Epoch.
+  @Deprecated('This constant will be removed in future without replacement.')
  Int64 get now;

+  double get nowNanoseconds;
}
```


```diff
/// DateTimeTimeProvider retrieves timestamps using DateTime.
class DateTimeTimeProvider implements TimeProvider {
  @override
  Int64 get now => Int64(DateTime.now().microsecondsSinceEpoch) * 1000;

+  @override
+  double get nowNanoseconds => DateTime.now().microsecondsSinceEpoch * 1.0;
}
```

```diff
class WebTimeProvider implements TimeProvider {
  /// The current time, in nanoseconds since Unix Epoch.
  ///
  /// Note that this time may be inaccurate if the executing system is suspended
  /// for sleep.  See https://github.com/open-telemetry/opentelemetry-js/issues/852
  /// for more information.
  @override
  Int64 get now => fromDOMHighResTimeStamp(window.performance.now());

+  @override
+  double get nowNanoseconds => window.performance.now();
}
```

yuzurihaaa:
hi @blakeroberts-wk . Can I get some feedback on https://github.com/Workiva/opentelemetry-dart/pull/207#discussion_r1981578703 if I got this right?

lib/src/sdk/logs/log_record.dart
yuzurihaaa: Hi @blakeroberts-wk , I pushed the changes with this new idea.

yuzurihaaa:
hi @blakeroberts-wk , can I know if there's any update? Or anyone else can help on the reviewing process.

yuzurihaaa:
hi @blakeroberts-wk , can I know if there's any update? Or @evanweible-wfcan help on the reviewing process?

yuzurihaaa:
hi @blakeroberts-wk , can I know if there's any update? Or @evanweible-wf  can help on the reviewing process?

lib/src/sdk/platforms/web/time_providers/web_time_provider.dart
```diff
@@ -51,4 +51,7 @@ class WebTimeProvider implements TimeProvider {
   /// for more information.
   @override
   Int64 get now => fromDOMHighResTimeStamp(window.performance.now());
+
+  @override
+  double get nowNanoseconds => window.performance.now();
```
blakeroberts-wk, line 56: This returns milliseconds, not nanoseconds

lib/src/sdk/time_providers/time_provider.dart
```diff
@@ -15,5 +15,8 @@ abstract class TimeProvider {
   static const int nanosecondsPerMillisecond = 1000000;
 
   /// The current time, in nanoseconds since Unix Epoch.
+  @Deprecated('This getter will be removed in future without replacement.')
   Int64 get now;
+
+  double get nowNanoseconds;
```
blakeroberts-wk, line 21: I think if we want to change this API, we should do something like `Duration get nowDuration;`.

We'd want to introduce new API and deprecate the current prior to 0.19.0 release. Then in 0.19.0, the return type of `now` is updated. And in 0.20.0, `nowDuration` is removed.

```dart
abstract class TimeProvider {
  /// The current time in nanoseconds since Unix Epoch.
  ///
  /// **Warning:** Return type will change to `Duration` in 0.19.0.
  /// Use [nowDuration] instead.
  @Deprecated(
    'Return type will change to `Duration` in 0.19.0. '
    'Use `nowDuration` for now, and migrate back to `now` after 0.19.0.'
  )
  Int64 get now;

  /// The current time as a [Duration] since Unix Epoch.
  ///
  /// **Warning:** Temporary API, will be removed in 0.20.0.
  /// Use this for intermediate migration of `now` prior to 0.19.0.
  Duration get nowDuration;
}
```

Then in 0.19.0, the deprecation of `now` will be removed when its type is updated and `nowDuration` will get a deprecation message.

The implementation of `nowDuration` would be something like the following:

```dart
class WebTimeProvider implements TimeProvider {
  static final Duration timeOrigin = Duration(
    milliseconds: (window.performance.timeOrigin ??
            window.performance.timing.navigationStart)
        .round(),
  );

  @override
  Int64 get now => fromDOMHighResTimeStamp(window.performance.now());

  @override
  Duration get nowDuration =>
      WebTimeProvider.timeOrigin +
      Duration(microseconds: (window.performance.now() * 1000).round());
}
```

```dart
class DateTimeTimeProvider implements TimeProvider {
  @override
  Int64 get now => Int64(DateTime.now().microsecondsSinceEpoch) * 1000;

  @override
  Duration get nowDuration =>
      Duration(microseconds: DateTime.now().microsecondsSinceEpoch);
}
```

lib/src/sdk/platforms/web/time_providers/web_time_provider.dart
yuzurihaaa: Hi @blakeroberts-wk , can I know with https://github.com/Workiva/opentelemetry-dart/pull/207#discussion_r2045084076 is this still valid?

yuzurihaaa:
Hi @robbecker-wf . Sorry that I missed your message, I've added the changes from review.

lib/src/sdk/time_providers/time_provider.dart
yuzurihaaa: updated

lib/src/sdk/logs/log_record_limit.dart
```diff
@@ -0,0 +1,17 @@
+// Copyright 2021-2022 Workiva.
+// Licensed under the Apache License, Version 2.0. Please see https://github.com/Workiva/opentelemetry-dart/blob/master/LICENSE for more information
+
+class LogRecordLimits {
+  final int _attributeCountLimit;
+  final int _attributeValueLengthLimit;
```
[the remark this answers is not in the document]
yuzurihaaa: updated

lib/src/sdk/logs/log_record.dart
```diff
@@ -0,0 +1,152 @@
...
+
+  final sdk.Resource _resource;
+
+  final sdk.TimeProvider _timeProvider;
+  final api.Context _context;
+  final sdk.LogRecordLimits logRecordLimits;
+  final DateTime? _timeStamp;
+  final DateTime? _observedTimestamp;
+
+  bool _isReadonly = false;
```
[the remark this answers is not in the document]
yuzurihaaa: removed. But I keep the write as is, and with this, it will always allow modifying.

lib/src/sdk/logs/log_record.dart
blakeroberts-wk: `window.performance.now()` returns milliseconds. Thinking on this further, let's go ahead and keep with the current paradigm using Int64 and we'll make the necessary breaking changes to use `double` for all APIs at sometime point in the future

lib/src/sdk/platforms/web/time_providers/web_time_provider.dart
blakeroberts-wk: It is not.

yuzurihaaa:
Hi @blakeroberts-wk , anything else that is pending from me to fix? In case if I missed something out.

dustinlessard-wf:
Hi @yuzurihaaa  
Thank you for taking the time to contribute to this project. We really appreciate the effort you put into opening this pull request.
I want to be transparent about the state of the repository: we have not been actively maintaining this codebase, and we do not currently have the bandwidth to properly review, test, and support external contributions at the standard they deserve.
For that reason, we’re going to close this pull request. This is not a reflection on the quality or value of your contribution. Rather, it reflects our current inability to responsibly vet and merge changes into the project.
We are also exploring a path to transfer ownership or stewardship of the project outside of our company, so that it can hopefully have a more active and sustainable future. Until that direction is clearer, we do not want to create a misleading impression that the repository is actively maintained or that we can support merged changes going forward.
Thank you again for your contribution and for your interest in improving the project. We’re sorry we were not able to engage with your pull request in the way it deserved.

PR #214, opened 2025-03-03
Fix trace and traceSync not creating root spans when newRoot is true

blakeroberts-wk:

`trace` and `traceSync` have a `newRoot` optional parameter that works by using `Context.root` to start the span. However, if there is a context with a span attached to the root zone, then using `Context.root` will make the span a child of the one held by the context attached to the root zone.

The `Tracer` defined by the SDK supports the `newRoot` optional parameter that results in appropriately configuring the span's parent span context to be invalid regardless of any span potentially visible via the given context.

However, the `Tracer` defined by the API does not support the `newRoot` optional parameter. Ideally, this interface would be updated to include the optional parameter. But this is a breaking change.

Therefore, to resolve the issue of `newRoot` not creating root spans, `trace` and `traceSync` type check the `tracer` to see if it is the SDK's `Tracer`. If so, the tracer's `newRoot` option is used. Otherwise, the API's `tracer` is used as usual.

## How Has This Been Tested?

The unit tests added in this PR were ran with and without the changes. Without the changes, the unit tests fail. With the changes, the unit tests pass.

## Checklist:

### DIFF
--- a/lib/src/api/open_telemetry.dart
+++ b/lib/src/api/open_telemetry.dart
@@ -6,6 +6,7 @@ import 'dart:async';
 import 'package:meta/meta.dart';
 
 import '../../api.dart' as api;
+import '../../src/sdk/trace/tracer.dart' as sdk show Tracer;
 import '../experimental_api.dart';
 import 'propagation/noop_text_map_propagator.dart';
 import 'trace/noop_tracer_provider.dart';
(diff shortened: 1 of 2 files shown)


blakeroberts-wk:
QA +1 CI is sufficient

aaronforest-wf (approved):

lib/src/api/open_telemetry.dart
```diff
@@ -102,12 +103,23 @@ Future<T> trace<T>(String name, Future<T> Function() fn,
   context ??= api.Context.current;
   tracer ??= _tracerProvider.getTracer('opentelemetry-dart');
 
-  final span = tracer.startSpan(name,
-      // TODO: use start span option `newRoot` instead
-      context: newRoot ? api.Context.root : context,
-      attributes: spanAttributes,
-      kind: spanKind,
-      links: spanLinks);
+  // TODO: use start span option `newRoot` instead
```
matthewnitschke-wk, line 106: Has this TODO essentially been resolved now? thoughts on removing it here and below?

matthewnitschke-wk (approved):

lib/src/api/open_telemetry.dart
blakeroberts-wk: No. The option is only on the sdk's implementation class of the api's interface which doesn't have the option. The next breaking release (0.16.0) will need to contain the change to add the new option to the interface that will allow us to remove this TODO and simplify the code down to just invoking the method on the given interface

blakeroberts-wk:
@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

Issue #218, opened 2025-04-14
How to set the timeout for request reporting?


StanleyCocos:
I want to set the HTTP request timeout, but currently I only see an option to pass an HTTP client.

PR #219, opened 2025-04-15
feat: add timeout

StanleyCocos:
close:  https://github.com/Workiva/opentelemetry-dart/issues/218

## Checklist:

### DIFF
--- a/lib/src/sdk/trace/exporters/collector_exporter.dart
+++ b/lib/src/sdk/trace/exporters/collector_exporter.dart
@@ -23,10 +23,11 @@ class CollectorExporter implements sdk.SpanExporter {
   final Uri uri;
   final http.Client client;
   final Map<String, String> headers;
+  final Duration? timeout;
   var _isShutdown = false;
 
   CollectorExporter(this.uri,
-      {http.Client? httpClient, this.headers = const {}})
+      {http.Client? httpClient, this.headers = const {}, this.timeout})
       : client = httpClient ?? http.Client();
 
   @override
@@ -58,8 +59,10 @@ class CollectorExporter implements sdk.SpanExporter {
 
     while (retries < maxRetries) {
       try {
-        final response = await client.post(uri,
-            body: body.writeToBuffer(), headers: headers);
+        final request =
+            client.post(uri, body: body.writeToBuffer(), headers: headers);
+        final response =
+            await (timeout != null ? request.timeout(timeout!) : request);
         if (response.statusCode == 200) {
           return;
         }
--- a/lib/src/sdk/trace/exporters/collector_exporter.dart
+++ b/lib/src/sdk/trace/exporters/collector_exporter.dart
@@ -23,12 +23,15 @@ class CollectorExporter implements sdk.SpanExporter {
   final Uri uri;
   final http.Client client;
   final Map<String, String> headers;
-  final Duration? timeout;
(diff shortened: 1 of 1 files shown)


lib/src/sdk/trace/exporters/collector_exporter.dart
```diff
@@ -23,10 +23,11 @@ class CollectorExporter implements sdk.SpanExporter {
   final Uri uri;
   final http.Client client;
   final Map<String, String> headers;
+  final Duration? timeout;
   var _isShutdown = false;
 
   CollectorExporter(this.uri,
-      {http.Client? httpClient, this.headers = const {}})
+      {http.Client? httpClient, this.headers = const {}, this.timeout})
```
blakeroberts-wk, line 30: Should we default to 10 seconds? I think Go, Java, and Python OTLP HTTP exporters default to 10 seconds iirc

StanleyCocos: Sure, but if the default is 10 seconds, then my request might be considered a significant change.

blakeroberts-wk: 10 seconds is a long time for a telemetry request that batches for 10 seconds by default. I think it's fine

blakeroberts-wk (dismissed):

lib/src/sdk/trace/exporters/collector_exporter.dart
```diff
@@ -23,11 +23,15 @@ class CollectorExporter implements sdk.SpanExporter {
...
   var _isShutdown = false;
 
-  CollectorExporter(this.uri,
-      {http.Client? httpClient, this.headers = const {}})
-      : client = httpClient ?? http.Client();
+  CollectorExporter(
+    this.uri, {
+    http.Client? httpClient,
+    this.headers = const {},
+    this.timeout = const Duration(seconds: 10),
```
blakeroberts-wk, line 33: I think we still need to support null to represent "no timeout", though it won't be the default

StanleyCocos: Making the timeout parameter optional doesn’t solve the issue, because when the user explicitly passes null from the outside, it still won’t take effect due to the default value.

To support setting it to null, we might need an additional flag to control the behavior, such as isTimeoutEnabled.

However, we need to consider whether this approach aligns with our expectations.

blakeroberts-wk: Ah. Damn. I think zero duration would be "off"?

blakeroberts-wk: Wait no that times out immediately doesn't it. And negative durations aren't supported. I suppose we could use a raw number that represents the timeout in milliseconds in which we construct the duration on the fly where negative numbers are used to represent "off"?

StanleyCocos: Good idea. I’ll try updating.

blakeroberts-wk (approved):

blakeroberts-wk:
@Workiva/release-management-p

StanleyCocos:
@blakeroberts-wk  Thank you.
May I ask if two approvals are required before merging? Also, I’d like to know if this feature will be included in the next release. Are there any plans for the next version?

